PR #1, opened 2018-08-09
It works! It works!!!

aaron-suarez:


db_setup/app/__init__.py
```diff
@@ -18,14 +18,26 @@ def import_resources():
     with open('resources.yml', 'r') as f:
         data = yaml.load(f)
 
+    resource_dict = {}
+    unique_resources = []
+
     for resource in data:
 
+        if resource['url'] not in resource_dict:
```
wimo7083: this is a linear search. doing
` if not resource.get('url', None):`
is going to be faster (hash speed lookup, and catches the exception by setting default value to `None`

PR #2, opened 2018-08-26
Create a resource if it doesn't exist, update a resource if it does

aaron-suarez:
There are still some optimizations to be done, but I'm starting this PR with something that works. We can continue to optimize before we actually merge it if you feel like that's necessary

### DIFF
--- a/db_setup/app/__init__.py
+++ b/db_setup/app/__init__.py
@@ -48,6 +48,19 @@ def import_resources():
         else:
             create_resource(resource)
 
+    try:
+        db.session.commit()
+    except exc.SQLAlchemyError as e:
+        db.session.rollback()
+        print('Flask SQLAlchemy Exception:', e)
+        template = "An SQLAlchemy exception of type {0} occurred. Arguments:\n{1!r}"
+        print(resource)
+    except Exception as e:
+        db.session.rollback()
+        print('exception', e)
+        template = "An exception of type {0} occurred. Arguments:\n{1!r}"
+        print(resource)
+
 def remove_duplicates(data):
     unique_resources = []
     resource_dict = {}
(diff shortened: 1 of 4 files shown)


db_setup/app/__init__.py
```diff
@@ -16,63 +17,137 @@
...
+        template = "An SQLAlchemy exception of type {0} occurred. Arguments:\n{1!r}"
+        print(resource)
+    except Exception as e:
+        db.session.rollback()
+        print('exception', e)
+        template = "An exception of type {0} occurred. Arguments:\n{1!r}"
+        print(resource)
+
+def update_resource(resource, existing_resource, langs, category):
+    # Return without updating if the resource is already up to date
```
wimo7083: should we do 
`
resource['languages'] = langs
`
and then change method:
`
match_resource(resource, existing_resource)
`

db_setup/app/__init__.py
```diff
@@ -16,63 +17,137 @@
...
+        print('Flask SQLAlchemy Exception:', e)
+        template = "An SQLAlchemy exception of type {0} occurred. Arguments:\n{1!r}"
+        print(resource)
+    except Exception as e:
+        db.session.rollback()
+        print('exception', e)
+        template = "An exception of type {0} occurred. Arguments:\n{1!r}"
+        print(resource)
+
+def match_resource(resource_dict, resource_obj, langs):
```
wimo7083: this comparison can be moved into the `Resource` class, using the `__eq__` magic method, or keep it as is. 


```
def match_resource(resource_dict, resource_obj):
    comparable_attributes = ['name', 'paid', 'notes', 'category', 'languages']
    
    return all( resource_dict.get(attr_type) == getattr(resource_obj, attr_type) for attr_type in comparable_attributes)
```


this can be done when you set the `__eq__` and `__hash__` methods for category also (like you did for languages).

db_setup/app/__init__.py
```diff
@@ -16,63 +17,137 @@
...
+        if existing_resource:
+            update_resource(resource, existing_resource, langs, category)
+        else:
+            create_resource(resource, langs, category)
 
+def remove_duplicates(data):
+    unique_resources = []
+    resource_dict = {}
+    for resource in data:
+        if not resource_dict.get(resource['url']):
```
wimo7083: If you encounter duplicate url you only want to use first one?

I'd also add some type of logging so you know when this happens. This seems like one of those cases that could turn into a really weird bug later when we don't know why some resources are not being shown but in the yaml.

db_setup/app/__init__.py
```diff
@@ -16,63 +17,137 @@
...
-            resource_dict[resource['url']] = True
-            unique_resources.append(resource)
+    # Step 3: Get existing entries from DB
+    try:
+        resources_list = Resource.query.all()
+        languages_list = Language.query.all()
+        categories_list = Category.query.all()
 
+        # Convert to dict for quick lookup
+        existing_resources = {r.url:r for r in resources_list}
```
wimo7083: This isn't bad if you need to preserve the actual item, which you are. You could give give each model class a method 
```
def key(self):
    return self.name  # or self.url
```
then here you'd call 
` {key(l) : l for l in languages_list } `

db_setup/app/__init__.py
```diff
@@ -16,63 +17,137 @@
...
+    for resource in data:
+        if not resource_dict.get(resource['url']):
+            resource_dict[resource['url']] = True
+            unique_resources.append(resource)
+    return unique_resources
+
+def get_category(resource, category_dict):
+    category = resource.get('category')
+
+    if category not in category_dict:
```
wimo7083: try to use the 
`category_dict.get(category, False)` 
pattern here since it's a constant lookup time instead of linear search.

db_setup/app/__init__.py
aaron-suarez: I like this, I'm definitely going to move this into the `Resource` class and define `__eq__` and `__hash__` for Resource and Category

db_setup/app/__init__.py
aaron-suarez: Good point, I'll add the logging so we know when it happens and can fix the yaml and run it again

db_setup/app/__init__.py
aaron-suarez: Ah, yes good call. This became obvious once I refactored to the point where I'm at. It was less obvious to do it this way with the original code. Now that you point it out, I'm going to implement that

db_setup/app/__init__.py
aaron-suarez: Funny, after implementing the logging, I found 2 duplicate resources somehow were still in there. Good call 👍

PR #3, opened 2018-09-07
Remove unnecessary dependencies from requirements.txt

aaron-suarez:
If we need any of these dependencies again, we can add them back in. @wimo7083 if you see anything that we're obviously going to need, point it out and I'll add it back in.


wimo7083:
I'll go ahead and merge this. If I find any issues I'll try and fix later. Thanks!

PR #17, opened 2018-10-07
Move app from db_setup to the root directory

aaron-suarez:
There was a point in time where we wanted to keep the `resources.yml` file and all of the associated data management files in a directory called `db_setup`. There was an access problem where we could only do this if we had the app in that directory as well. Now that we have the `import_resources` function behind a click command, I don't think it makes sense to have a db_setup folder at all.

Open to discussion about a better way to structure it, but this PR contains the basics of my idea. Everything appears to work fine on my machine. Tested the following:

- Importing resources with click command
- Pulling resources via `GET /resources`
- Downgrading the db with `flask db downgrade`
- Migrating the db with `flask db migrate`
- Upgrading the db with `flask db upgrade`

Next steps:
- We need to break out the `populate_db` click command into it's own file so that it's not in __init__.py.
- We need to break out the route for `GET /resources` into it's own `routes.py` file, where all of the routes will live


wimo7083:
Before:

```bash
.
├── db_setup
│   ├── app
│   │   ├── app.py
│   │   ├── health_check.py
│   │   ├── __init__.py
│   │   └── models.py
│   ├── configs.py
│   ├── migrations
│   │   ├── alembic.ini
│   │   ├── env.py
│   │   ├── README
│   │   ├── script.py.mako
│   │   └── versions
│   │       ├── 158d296b23b8_.py
│   │       ├── 514a36eb1161_.py
│   │       ├── 579c613eff61_initial_add.py
│   │       ├── 59a6f6552b5a_.py
│   │       └── 9dd599c1495d_set_default_allow_languages_nullable.py
│   ├── resources.yml
│   └── run.py
├── LICENSE
├── README.md
└── requirements.txt
```

wimo7083:
After:
```bash
.
├── app
│   ├── app.py
│   ├── health_check.py
│   ├── __init__.py
│   └── models.py
├── configs.py
├── db_setup
├── LICENSE
├── migrations
│   ├── alembic.ini
│   ├── env.py
│   ├── README
│   ├── script.py.mako
│   └── versions
│       ├── 158d296b23b8_.py
│       ├── 514a36eb1161_.py
│       ├── 579c613eff61_initial_add.py
│       ├── 59a6f6552b5a_.py
│       └── 9dd599c1495d_set_default_allow_languages_nullable.py
├── README.md
├── requirements.txt
├── resources.yml
└── run.py
```

PR #19, opened 2018-10-12
Restructure the directory to something that makes more sense

aaron-suarez:
This breaks out the command for populating the db into its own file. It also breaks out the routes into their own file and makes them `/api/v1/{route}`


aaron-suarez:
New structure looks something like this

```
.
├── LICENSE
├── README.md
├── app
│   ├── __init__.py
│   ├── api
│   │   ├── __init__.py
│   │   └── routes.py
│   ├── cli.py
│   ├── health_check.py
│   └── models.py
├── configs.py
├── requirements.txt
├── resources.yml
├── run.py
```

PR #21, opened 2018-10-17
Add pagination for the get_resources route

### LINKED ISSUE #15: Implement pagination
When pulling all the resources with `GET /resources`, we need to limit results so it doesn't take forever to return the JSON. Right now in my local setup with a remote database, it takes about 3 minutes to get all the resources, which is way too long. I think returning results 20 at a time might be a good number to start with, but if anyone has any other opinions on the number of results, feel free to comment on this issue and we can discuss.

aaron-suarez:
Fixes #15 

Pagination is implemented using Flask-SQLAlchemy's built-in paginator. The specific page is requested using a query param called `page`. The request looks like this:
```
GET api/v1/resources?page=1
```
The number of resources per page is included in the Config class, so we can easily change it later. I imagine that eventually we will add a `CATEGORIES_PER_PAGE` and `LANGUAGES_PER_PAGE` as well, for the `GET api/v1/categories` and `GET api/v1/languages` routes.

### DIFF
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -7,6 +7,7 @@ from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
 from app.api import bp
 from app.models import Resource, Language
 from app import Config
+from app.utils import Paginator
 
 
 # Routes
@@ -48,10 +49,7 @@ def get_resource(id):
 
 def get_resources():
     try:
-        page = request.args.get('page', 1, type=int)
-        page_size = request.args.get('page_size', Config.RESOURCES_PER_PAGE, type=int)
-        if page_size > Config.RESOURCE_MAX_PAGE_SIZE: page_size = Config.RESOURCE_MAX_PAGE_SIZE
-        resource_paginator = Resource.query.paginate(page, page_size, False)
+        resource_paginator = Paginator(Config.RESOURCE_PAGINATOR, Resource, request)
         resource_list = [resource.serialize for resource in resource_paginator.items]
 
     except Exception as e:
@@ -66,10 +64,7 @@ def get_languages():
     languages = {}
 
     try:
-        page = request.args.get('page', 1, type=int)
-        page_size = request.args.get('page_size', Config.LANGUAGES_PER_PAGE, type=int)
-        if page_size > Config.LANGUAGES_MAX_PAGE_SIZE: page_size = Config.LANGUAGES_MAX_PAGE_SIZE
-        language_paginator = Language.query.paginate(page, page_size, False)
+        language_paginator = Paginator(Config.LANGUAGE_PAGINATOR, Language, request)
         language_list = [language.serialize for language in language_paginator.items]
 
     except Exception as e:
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -1,10 +1,12 @@
 from traceback import print_tb
 
+from flask import request
 from flask import jsonify
 from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
 
 from app.api import bp
 from app.models import Resource, Language
+from app import Config
 
 
 # Routes
@@ -47,14 +49,16 @@ def get_resource(id):
 def get_resources():
     resources = {}
     try:
-        resources = Resource.query.all()
+        page = request.args.get('page', 1, type=int)
+        resource_paginator = Resource.query.paginate(page, Config.RESOURCES_PER_PAGE, False)
+        resource_list = [resource.serialize for resource in resource_paginator.items]
 
     except Exception as e:
         print_tb(e.__traceback__)
         print(e)
 
     finally:
-        return jsonify([single_resource.serialize for single_resource in resources])
+        return jsonify(resource_list)
 
 
 def get_languages():
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -49,7 +49,9 @@ def get_resource(id):
 def get_resources():
     try:
         page = request.args.get('page', 1, type=int)
-        resource_paginator = Resource.query.paginate(page, Config.RESOURCES_PER_PAGE, False)
+        page_size = request.args.get('page_size', Config.RESOURCES_PER_PAGE, type=int)
+        if page_size > Config.RESOURCE_MAX_PAGE_SIZE: page_size = Config.RESOURCE_MAX_PAGE_SIZE
+        resource_paginator = Resource.query.paginate(page, page_size, False)
         resource_list = [resource.serialize for resource in resource_paginator.items]
 
     except Exception as e:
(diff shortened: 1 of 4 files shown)


app/api/routes.py
```diff
@@ -17,25 +19,28 @@ def languages():
 
 
 def get_resources():
-    resources = {}
     try:
-        resources = Resource.query.all()
+        page = request.args.get('page', 1, type=int)
+        resource_paginator = Resource.query.paginate(page, Config.RESOURCES_PER_PAGE, False)
```
wimo7083: Can you add an optional query parameter to override this `RESOURCES_PER_PAGE` not to exceed a maximum value?

app/api/routes.py
```diff
@@ -17,25 +19,28 @@ def languages():
...
+        resource_list = [resource.serialize for resource in resource_paginator.items]
 
     except Exception as e:
         print_tb(e.__traceback__)
         print(e)
-
+        resource_list = []
     finally:
-        return jsonify([single_resource.serialize for single_resource in resources])
+        return jsonify(resource_list)
```
wimo7083: We'll eventually want to start returning more than a json array. But this is good for now.

app/api/routes.py
```diff
@@ -17,25 +19,28 @@ def languages():
...
 
 
 def get_languages():
     try:
-        languages = Language.query.all()
+        page = request.args.get('page', 1, type=int)
+        language_paginator = Language.query.paginate(page, Config.LANGUAGES_PER_PAGE, False)
```
wimo7083: Same as above for overriding value.

app/models.py
```diff
@@ -99,6 +99,14 @@ class Language(db.Model):
     id = db.Column(db.Integer, primary_key=True)
     name = db.Column(db.String, unique=True, nullable=False)
 
+    @property
```
wimo7083: we have a `serialize_language` method on `Resources` how do you see that tieing into this?

configs.py
```diff
@@ -22,3 +22,5 @@ class Config:
     SQLALCHEMY_TRACK_MODIFICATIONS = False
     SQLALCHEMY_DATABASE_URI = os.environ.get('SQLALCHEMY_DATABASE_URI') or alternative_db
     SQL_LITE_DEFAULT = False or (alternative_db != None)
+    RESOURCES_PER_PAGE = 20
```
wimo7083: 1500/20 = 75 pages.
50?/20 = 2.5 pages. 

bigger default

app/api/routes.py
aaron-suarez: Ooh, good call. Should we call it `max_num`? So it will be like `api/v1/resources?page=2&max_num=10`

app/api/routes.py
aaron-suarez: Feel free to open an issue so that someone can jump on fixing that for all the routes. I'm not sure what else we ought to be returning. We're not rendering views from the API, are we?

configs.py
aaron-suarez: I chose somewhat random values. I'm open to suggestions on better defaults.

app/api/routes.py
wimo7083: Slack uses 
`limit`
twilio uses 
`PageSize`
I actually like `page_size` if you can find one that's more standard or better we go with that one. Don't want to bikeshed naming too much.

app/models.py
aaron-suarez: Well, the thing about it is that it's a Many To Many relationship. So, in some contexts, we want a list of languages (as strings), and in other contexts we want the IDs included in the data. 

So, let's suppose you get all the languages (including IDs). Then you can take an ID and send a request to get all the resources associated with that. Or all the categories, etc..

Now, let's suppose you get a certain resource. You just want the languages associated with it, you don't necessarily need the IDs of those languages, do you?

So they serve different purposes.

app/api/routes.py
wimo7083: Things like status, or total number of resources, or number of total pages, or some type of cache reference. 

Like all these guys:
https://api.slack.com/methods/users.list

aaron-suarez: Oh, right. Yeah, we'll probably want to add some of those.

app/api/routes.py
```diff
@@ -35,7 +37,9 @@ def get_resources():
 def get_languages():
     try:
         page = request.args.get('page', 1, type=int)
-        language_paginator = Language.query.paginate(page, Config.LANGUAGES_PER_PAGE, False)
+        page_size = request.args.get('page_size', Config.LANGUAGES_PER_PAGE, type=int)
```
wimo7083: I like the use of type setting here. If someone passes in `?page=bcd2` we'll crash. 

What do you think should happen?

aaron-suarez: Either we should use the default if something unexpected is encountered, or we should return an error code that says "invalid_value" with a message that explains that `bcd2 is invalid` or something. But I like the former, because if the client side programmer does something wrong, the user doesn't suffer because of it.

What do you think?

aaron-suarez: Also, I just passed in `bcd2` and it didn't crash, it returned the default number of resources. We can conduct more extensive tests, but I suspect that the behavior we're getting is actually what I described as my preference above.

heavysaturn:
there's a bit of repetition when the pagenator is set up. perhaps we could make this more DRY by making the paginator a class and doing the setup in an `__init__` method, then just importing the class and calling its constructor with something like `paginator = Paginator("language")` to tell it which config to load. or we could pass it a config dict... or even a 3.7 dataclass!

heavysaturn:
I think it's inelegant to have five lines of the same config in every endpoint with pagination.

I've been thinking a bit about how to improve on that, design-wise, and the best I can come up with is making a Paginator class that takes a 3.7 dataclass and a model, and handles the initialization.

for example:

```py
# configs.py
from dataclasses import dataclass

@dataclass
class PaginatorConfig:
    per_page: int = 20
    max_page_size: int = 100

class Config:
    SQLALCHEMY_DATABASE_URI = os.environ.get('SQLALCHEMY_DATABASE_URI') or alternative_db
    SQL_LITE_DEFAULT = False or (alternative_db != None)
    RESOURCE_PAGINATOR = PaginatorConfig()
    LANGUAGE_PAGINATOR = PaginatorConfig(
        per_page=40,  # we can pass in changes like this, if we want them different
    )


# in utils.py or something like that
class Paginator:
    def __init__(self, configuration, model):
        self.configuration = configuration
        self.model = model

        # request should still work, since it's a context local 
        self.page = request.args.get('page', 1, type=int)  
        self.page_size = request.args.get('page_size', configuration.per_page, type=int)
        
        if self.page_size > configuration.max_page_size: 
            self.page_size = configuration.max_page_size
        
    @property
    def items(self):
        return self.model.query.paginate(self.page, self.page_size, False)
        

# in routes.py
from utils import Paginator

resource_paginator = Paginator(Config.RESOURCE_PAGINATOR, Resource)
resource_list = [resource.serialize for resource in resource_paginator.items]
```

This is just a suggestion, of course, but something like that would make it easy to add pagination to any number of endpoints.

aaron-suarez:
Yes, I also had a bad taste in my mouth about the code repetition, but I wasn't entirely sure how to abstract it away from those methods. I'm going to play with your suggestion and see if I can get something working.

aaron-suarez:
@wimo7083 Please review the changes one more time and if everything checks out, let's get this merged!

@heavysaturn Thanks for the POC code. There were a few changes that I needed to make (had to pass in the request because it turned out that it was not in scope, had to edit items slightly). Overall, you made the job much easier than if I were trying to figure out all of the syntax on my own, so thanks for that.

Issue #14, opened 2018-10-07
Routes needed


wimo7083:
I'm going to be lazy and put these here in one. Think of them each as an issue. 

- [x] GET /resources (get all resources, probably want some pagination for this)
- [ ] GET /resources/{id} (get a specific resource)
- [ ] GET /categories (get the list of available categories)
- [ ] GET /categories/{category} (get the resources for a named category, probably want pagination for this)
- [x] GET /languages (get the list of available languages)
- [ ] GET /languages/{language} (get the resources for a named language, probably want pagination for this)
- [ ] POST /resources (create a new resource)
- [ ] PUT /resources/{id} (update an existing resource)

wimo7083:
After reading about api practices:
1. We are using nouns and not verbs 
2. We are using plural nouns
3. Using the HTTP status to correspond to the action
4. Pagination with default values and a way to reference  page. 
5. version the api `api/v1/resources..`

aaron-suarez:
I've decided that I don't like the name of the route `GET /languages/{language}` to get all of the resources associated with a language, nor `GET /categories/{category}` for all the resources for a category.

I think that naming it as such makes you expect that it would return a single `language` or `category` with that name, but that's not what it returns at all. I would think it should be `GET /resources/{language}` and `GET /resources/{category}`.

The problem with that, however, is that they are both strings, so how would we know whether the person intended for their string to be a `language` or a `category`?

I think we should have a conversation about this, even if we end up sticking with the original naming scheme.

wimo7083:
`GET /resources?category=value?language=value2`?

aaron-suarez:
if you meant `GET /resources?category=value&language=value2`, I think that's a great solution. Plus, it's named explicitly, and you can filter for both category and language as you like

aaron-suarez:
Closing this issue in favor of specific issues that I just opened up

PR #26, opened 2018-10-18
Add an endpoint for getting a specific resource

heavysaturn:
This adds an endpoint that allows you to get a specific resource by id. This resolves one of the issues in #14

### DIFF
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -1,19 +1,47 @@
 from traceback import print_tb
 
 from flask import jsonify
+from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
 
 from app.api import bp
 from app.models import Resource, Language
 
 
+# Routes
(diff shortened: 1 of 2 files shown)


app/api/routes.py
```diff
@@ -1,21 +1,49 @@
...
+        print_tb(e.__traceback__)
+        print(e)
+
+    except NoResultFound as e:
+        print_tb(e.__traceback__)
+        print(e)
+
+    finally:
+        if resource:
+            return jsonify([resource.serialize])
```
aaron-suarez: I'm not sure if it makes sense to return it as a list that always contains one item. Should we just do `return jsonify(resource.serialize)` and `return jsonify({})`?

aaron-suarez:
This endpoint should only ever return one resource, yeah? This does not need to be paginated.

aaron-suarez:
Nice. Bonus points for fixing the typo.

PR #35, opened 2018-10-18
Add travis file for running pylint continuously

aaron-suarez:
We should add travis to run pylint before we merge anything. This fixes #12

Issue #16, opened 2018-10-07
Implement tests


aaron-suarez:
We have no unit tests. There is no way of verifying that when changes are made that it doesn't break something else. It might not be a bad idea to implement the [unittest framework](https://docs.python.org/3/library/unittest.html), but if there's a more flask-specific solution, that would work as well.

wimo7083:
Pytest is frickin awesome, if you'd like to see how we did it with flask in the other repo you can view here:
https://github.com/OperationCode/operationcode_pyback/tree/master/tests

heavysaturn:
`unittest` is what Miguel Grinberg uses in his [fairly well known Flask books and tutorials](https://blog.miguelgrinberg.com/post/the-flask-mega-tutorial-part-vii-unit-testing-legacy), and it would do the job. 

There's also the [Flask-Testing extension](https://pythonhosted.org/Flask-Testing/), which provides some nice high level features like abstracting certain common assertions.

Lastly, there's `pytest`, which is [what the official Flask tutorial](http://flask.pocoo.org/docs/1.0/testing/) suggests.

These are all viable options, though. The second option perhaps has some extra bells and whistles and lends itself to splitting the tests into multiple files, so I think that would be my choice, but really any of these would be fine.

wimo7083:
I know unit tests are difficult to setup at first. But we really need to prioritize this. 

I'm going to make the call to use [pytest](https://pytest-flask.readthedocs.io/en/latest/)

wimo7083:
If anybody wants to see how we started tests on pyback with pytest here's an example.
https://github.com/OperationCode/operationcode_pyback

I'd also like discussion on what unit tests we should add first. Do we add tests for ensuring the environment is ready, or do we start putting our routes in tests?

heavysaturn:
Testing all the routes is a good place to start, in my opinion.

aaron-suarez:
I think we should test the cli and routes since that's the majority of our business logic. Once we get the testing framework set up, we can add some issues with some missing tests and let newbies help out with that effort

aaron-suarez:
The testing framework has been implemented. Who wants to write some tests?

PR #36, opened 2018-10-18
Add and fix linting

### LINKED ISSUE #12: No linting
We should include a PEP8 linter to make sure we format things correctly.

hugovk:
* Fixes #12 
* Includes and closes #35
* Switch from `pylint` to `flake8` linting, which includes `pyflakes` (checks for errors) and `pycodestyle` (formerly called `pep8`)
* Fix linting errors, including running `isort`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,11 @@
+language: python
+python:
+  - 3.6
+# command to install dependencies
+install:
+  - pip install -r requirements.txt
+# command to run tests
+script:
+  - pylint app
+# TODO: implement pytests then delete this line and uncomment the next
+#  - pytest
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,6 @@ install:
   - pip install -r requirements.txt
 # command to run tests
 script:
-  - pylint app
+  - flake8 app --statistics --count
 # TODO: implement pytests then delete this line and uncomment the next
 #  - pytest
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -7,9 +7,9 @@ from flask_sqlalchemy import SQLAlchemy
 
 try:
     assert version_info >= (3, 7, 0)
-except AssertionError as e:
-    print('Warning Current python version not supported')
-    print('Current python version: {version}'.format(version=version_info))
+except AssertionError:
+    print('Warning Current Python version not supported')
+    print('Current Python version: {version}'.format(version=version_info))
     exit(1)
 
 
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -9,7 +9,7 @@ try:
     assert version_info >= (3, 7, 0)
 except AssertionError:
     print('Warning Current Python version not supported')
-    print('Current Python version: {version}'.format(version=version_info))
+    print(f'Current Python version: {version_info}')
     exit(1)
 
 
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,16 +1,17 @@
+from sys import version_info
+
+from configs import Config
 from flask import Flask
 from flask_migrate import Migrate
 from flask_sqlalchemy import SQLAlchemy
-from sys import version_info
 
 try:
-    assert version_info >= (3,7,0)
+    assert version_info >= (3, 7, 0)
 except AssertionError as e:
     print('Warning Current python version not supported')
     print('Current python version: {version}'.format(version=version_info))
     exit(1)
-    
-from configs import Config
+
 
 db = SQLAlchemy()
 migrate = Migrate()
--- a/app/api/__init__.py
+++ b/app/api/__init__.py
@@ -1,5 +1,4 @@
+from app.api import routes
 from flask import Blueprint
 
 bp = Blueprint('api', __name__)
-
-from app.api import routes
--- a/app/api/__init__.py
+++ b/app/api/__init__.py
@@ -1,4 +1,3 @@
-from app.api import routes
 from flask import Blueprint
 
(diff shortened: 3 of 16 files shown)


heavysaturn:
did you forget to `git add` the tox.ini file?

hugovk:
No, there is no tox.ini file needed. 

This is for linting on the CI.

See https://github.com/OperationCode/resources_api/issues/16 for adding unit tests.

heavysaturn:
flake8 comes with `flake8 --install-hook git`, allowing you to add a flake8 pre-commit hook in a one-liner. This means you can lint your local stuff automatically before you commit it, That way you won't be forced to make additional commits just to fix linting errors discovered by the CI. This is a far better workflow than just leaving it to the CI.

If we're linting locally, stuff like `exclude=.venv` will be important. It's not unlikely that `max_line_length` should be changed and that certain linting codes should be ignored (B311 and S311 are notable examples, these will not allow you to do `import random`). 

For these reasons, I think a `tox.ini` would be a good inclusion

.travis.yml
```diff
@@ -0,0 +1,11 @@
...
+python:
+  - 3.6
+# command to install dependencies
+install:
+  - pip install -r requirements.txt
+# command to run tests
+script:
+  - flake8 app --statistics --count
+# TODO: implement pytests then delete this line and uncomment the next
+#  - pytest
```
aaron-suarez: Nit: When I initially committed this, I was missing the newline from the end of the file. I'd appreciate it if you'd please pick up my slack and stick that in 😉

hugovk: Added :)

app/__init__.py
```diff
@@ -1,16 +1,17 @@
...
 
 try:
-    assert version_info >= (3,7,0)
-except AssertionError as e:
-    print('Warning Current python version not supported')
-    print('Current python version: {version}'.format(version=version_info))
+    assert version_info >= (3, 7, 0)
+except AssertionError:
+    print('Warning Current Python version not supported')
+    print(f'Current Python version: {version_info}')
```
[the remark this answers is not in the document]
hugovk: Reverted!

hugovk:
> 1. Install instructions for someone who is new to coding to do this linting

Added to Development Notes in the README.

> 2. Incorporate changes Leon suggested (in some form).

I'm not so sure how best to do tox.ini and the pre-commit hook. Could someone else take that up in a separate PR?

I've added a `.flake8` config file. How does that look?

README.md
```diff
@@ -45,3 +45,10 @@ If you make changes to the models.py or other schemas, you need to run a migrati
 1. Set the FLASK_APP environment variable `export FLASK_APP=run.py` or `ENV:FLASK_APP = "run.py"`
 2. Run the migration `flask db migrate`
 3. Upgrade to the latest migration `flask db upgrade`
+
+Before committing, please lint your code:
+
+```sh
+pip install --upgrade flake8
```
aaron-suarez: If flake8 is installed via virtualenv, is upgrading it necessary? Why do we need to run this command before running `flake8 app`?

aaron-suarez: And to clarify, I'm legitimately asking because I don't know, not because I think we should remove this line. I'm sure it's fine, I just want to know why it's necessary.

hugovk: It's not necessary to run it right before `flake8`, as long as it's been installed somehow first.

Some people might not install via virtualenv (I haven't for this PR). And it might be fine without the upgrade, but I generally include it in instructions to ensure the tools are kept up-to-date.

We could remove the upgrade if you prefer.

aaron-suarez: No, I like it as is. I just assumed that anyone who gets to the bottom of the README would have followed the instructions and used virtualenv. But if they're knowledgable enough to know what kind of difference it makes, then they can adjust their command accordingly.

This seems like a good default for those who are kind of blindly following instructions and cargo culting trying to figure this stuff out.

heavysaturn:
The pre-commit hook comes with flake8, and you don't need to do anything to have that feature. You could perhaps mention in the documentation that `flake8 --install-hook git` will magically make flake8 run before any commit you try to make, and that if flake8 fails, it will stop the commit. This way they will not be allowed to commit code that does not pass the linter. (It is possible to bypass by doing `git commit --no-verify`, though)

As for the `tox.ini`, the `.flake8` is the same thing, so it's fine that you've done it with that filename instead. 

Here's the default config I usually have in my projects

```
[flake8]
max-line-length=120
exclude=.venv
ignore=B311,W503,E226,S311
import-order-style=pycharm
```

This assumes that the following flake8 plugins are installed:
- flake8-bugbear
- flake8-string-format
- flake8-todo
- flake8-import-order
- flake8-bandit

I will need @aaron-suarez or @wimo7083 to chime in on which of these we want for this project, and if you include them in the `requirements.txt` file you will also need to install them and do another round of linting and error correction, because they will probably discover some errors you haven't addressed yet.

Alternatively, I'd be happy to do all of the above in a separate PR, in which case this can just be merged because it looks pretty good.

aaron-suarez:
I like the PR as is, I would like to get it merged so we can rebase all of our existing PRs and move on. I would like to incorporate flake8-bandit for sure, we can look at the other ones. Since @heavysaturn is willing to do that work in a different PR and appears to be more familiar with these types of configuration files, he can do that PR.

@wimo7083, you had requested changes, and as far as I can tell they are incorporated. Can we go ahead and merge this?

aaron-suarez:
Probably need to rebase to fix the conflicts from merging #26... sorry about that

PR #38, opened 2018-10-20
Added endpoint for updating a resource

### LINKED ISSUE #34: Need PUT /resource/{id} Route
This endpoint should update a resource based on the id passed in and JSON form data given in the request.

*Required attributes*:
`id` - Integer. The id of the resource to update.

*Optional attributes*:
`category` - A string category. Create a new category if necessary, switch the association of this resource to the new category
`languages` - An array of strings representing each language associated with this resource. Endpoint should add any new languages to the languages table and associate this resource with any existing languages that are specified  
`name` - A string that's the name of the resource  
`url` - A string that is a valid URL. (Ideally, the endpoint would validate this URL by performing a GET request and ensuring it receives a 200 response, but that's not really necessary for the MVP, just something to think about)
`paid` - Boolean. True if the resource requires payment to use. False if it's a free resource
`notes` - String. Any notes that are relevant to the resource

michizhou:
Added code for a new endpoint that is designed to update a resource based on the id passed in and any optional JSON form data parameters given in a PUT request. Uses helper function to update resource with parameters from the request. Closes #34.

### DIFF
--- a/app/api/__init__.py
+++ b/app/api/__init__.py
@@ -1,8 +1,6 @@
 from flask import Blueprint
-from flask_sqlalchemy import SQLAlchemy
 
 bp = Blueprint('api', __name__)
-db = SQLAlchemy()
 
 # We need to import the routes here so they will
 # bind to the blueprint before the blueprint is registered.
--- a/app/api/__init__.py
+++ b/app/api/__init__.py
@@ -1,3 +1,5 @@
 from flask import Blueprint
+from flask_sqlalchemy import SQLAlchemy
 
 bp = Blueprint('api', __name__)
+db = SQLAlchemy()
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -99,6 +99,12 @@ def set_resource(id, param_list, args):
         if resource:
             for position in range(len(param_names)):
                 resource[param_names[position]] = param_list[position]
+            try:
+                db.session.commit()
+            except exc.SQLAlchemyError as e:
+                db.session.rollback()
+                print('Flask SQLAlchemy Exception:', e)
+                print(resource)
             return jsonify(resource.serialize)
         else:
             return jsonify({})
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -18,7 +18,7 @@ def resources():
 
 @bp.route('/resources/<int:id>', methods=['GET', 'PUT'])
 def resource(id, category=None, languages=[], name=None, url=None,
-                 paid=False, notes=None):
+             paid=False, notes=None):
     if request.method == 'GET':
         return get_resource(id)
     elif request.method == 'PUT':
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -16,24 +16,23 @@ def resources():
     return get_resources()
 
 
-@bp.route('/resources/<int:id>', methods=['GET'])
-def resource_get(id):
-    return get_resource(id)
+@bp.route('/resources/<int:id>', methods=['GET', 'PUT'])
+def resource(id, params="", category=None, languages=[], name=None, url=None,
+                 paid=False, notes=None):
+    if request.method == 'GET':
+        return get_resource(id)
+    elif request.method == 'PUT':
+        param_list = [category, languages, name, url, paid, notes]
+        param_names = ['category', 'languages', 'name', 'url', 'paid', 'notes']
+        for index in range(len(param_names)):
+            if request.args.get(param_names[index]):
+                param_list[index] = param_names[index]
+        return set_resource(id, param_list)     
 
 
 @bp.route('/languages', methods=['GET'])
 def languages():
     return get_languages()
-
-
-@bp.route('/resources/<int:id>/<string:params>', methods=['PUT'])
-def resource_set(id, params="", category=None, languages=[], name=None, url=None,
-                 paid=False, notes=None):
-    param_list = [category, languages, name, url, paid, notes]
-    optional_params = params.split('/')
-    for index in range(len(optional_params)):
-        param_list[index] = optional_params[index]
-    return set_resource(id, param_list)
 
 
 # Helpers
@@ -86,7 +85,7 @@ def get_languages():
 def set_resource(id, param_list):
     resource = None
     try:
-        resource = get_resource(id)
+        resource = Resource.query.get(id)
 
     except MultipleResultsFound as e:
         print_tb(e.__traceback__)
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -24,7 +24,15 @@ def resource(id):
 @bp.route('/languages', methods=['GET'])
 def languages():
     return get_languages()
-
+    
+@bp.route('/resources/<int:id>/<string:params>', methods=['PUT'])
+def resource(id,params="",category=None,languages=[],name=None,url=None,paid=False,notes=None):
+    param_list = [category,languages,name,url,paid,notes]
+    old_resource = get_resource(id)
+    optional_params = params.split('/')
+    for index in range(len(optional_params)):
+        param_list[index] = optional_params[index]
+    return set_resource(id,param_list)
 
 # Helpers
 def get_resource(id):
@@ -61,7 +69,6 @@ def get_resources():
 
 
 def get_languages():
-
     try:
         language_paginator = Paginator(Config.LANGUAGE_PAGINATOR, Language, request)
         language_list = [language.serialize for language in language_paginator.items]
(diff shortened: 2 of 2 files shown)


app/api/routes.py
```diff
@@ -72,3 +80,29 @@ def get_languages():
         language_list = []
     finally:
         return jsonify(language_list)
+
+
+def set_resource(id, param_list):
+    resource = None
+    try:
+        resource = Resource.query.get(id)
```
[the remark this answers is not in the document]
michizhou: While I agree it might be nice to abstract out, after looking over the code for updating/creating a resource in the CLI I am afraid that it might not be wise to try abstracting out the code since the CLI resources are pulled from the current DB session, with all their field values stored in dynamic dictionaries rather than setting explicit values passed by users.

There is also the issue with how the CLI resources only have some of their fields ('category' and 'languages') updated, which makes refactoring difficult since in this case all resource fields are being updated with information from user input. For now I will just edit the Route code to have it properly update/create a resource while leaving the CLI code as it is.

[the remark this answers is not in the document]
aaron-suarez: Yeah, those are good points, perhaps we can't really share that code easily

michizhou:
@aaron-suarez In my latest commit, in which I addressed the concerns of your review, the build seems to be failing due to some "invalid syntax" error that seems to be on a line of the code that does not exist. Do you know why this error may be coming up? Please let me know.

app/api/routes.py
```diff
@@ -16,9 +16,14 @@ def resources():
     return get_resources()
 
 
-@bp.route('/resources/<int:id>', methods=['GET'])
-def resource(id):
-    return get_resource(id)
+@bp.route('/resources/<int:id>', methods=['GET', 'PUT'])
+def resource(id, category=None, languages=[], name=None, url=None,
```
aaron-suarez: So I understand that `id` is being passed into this method because it's in the route itself. What is the purpose of passing in `category`, `languages`, etc? All of that info comes from `request.args`, right? These are unnecessary in the method signature, aren't they?

app/api/routes.py
```diff
@@ -72,3 +76,29 @@ def get_languages():
...
+        print(e)
+
+    except NoResultFound as e:
+        print_tb(e.__traceback__)
+        print(e)
+
+    finally:
+        if resource:
+            for position in range(len(param_names)):
+                resource[param_names[position]] = param_list[position]
```
aaron-suarez: After this happens, it needs to be saved in the database with `db.session.commit()`, which should go inside a try block.

app/api/routes.py
michizhou: The other parameters after `id` (i.e. `category`, `languages`, etc) are default values that serve to provide a field with an appropriate "empty" value if no user input is specified for that field. I was just concerned that if I simply return `request.args` for all fields then certain fields (i.e. `languages` and `paid`) would have `None` as their default values, rather than an empty list or False boolean for their defaults (which I think are more appropriate).

app/api/routes.py
michizhou: What is the error type I should specify for the `except` clause (i.e. something similar to `MultipleResultsFound` or `NoResultFound`)?

aaron-suarez: I would just catch all SQLAlchemy errors like we do here https://github.com/OperationCode/resources_api/blob/master/app/cli.py#L52

app/api/routes.py
aaron-suarez: Fair enough. Wouldn't it be better to just declare them as variables inside the function rather than pass them into the function? I just think it's hard to see how these are used when they're declared like this.

aaron-suarez:
Have you tried `from app import db`?

michizhou:
Thanks for the advice! Everything seems to be working now, with the build passing. Feel free to let me know if I should make any more changes to `routes.py` or any other files.

app/api/__init__.py
```diff
@@ -1,6 +1,8 @@
 from flask import Blueprint
+from flask_sqlalchemy import SQLAlchemy
 
 bp = Blueprint('api', __name__)
+db = SQLAlchemy()
```
aaron-suarez: Are these lines necessary if you're importing `db` from `app`?

michizhou: I suppose not. The only reason I included that line was because I assumed it was needed to define `db`, which would then be referenced by `routes.py`.

aaron-suarez:
I think that unlike JavaScript, Python objects _must_ use dot syntax for attributes

```python
resource = Resource()

# this works
resource.name = "name"

# this doesn't work
resource['name'] = "name"
```

I think that you're going to have to rethink how you're looping through the array of param names. Probably easier to just list them out and assign them one by one rather than in a for loop. Maybe there's a way to do it using your loop, but I'm not sure.

michizhou:
I actually did not try it out, since I do not have the proper testing environment set up to do so and assumed that I could use the code in `cli.py` as a guide. Should I just forget about using default values for the resource parameters and just use `request.args`? Or should I switch from using dict notation (i.e. `resource['category']`) to dot notation (i.e. `resource.category`) to assign resource parameters?

aaron-suarez:
We used dot notation in `cli.py`, so I believe that would work. 

Also, yeah, I was kind of confused from the beginning why you would need any defaults for updating a resource. For creating a resource, you'll need default values for that stuff (but I think the model actually creates some of those defaults for you. e.g., `paid` defaults to `False`). 

But for updating a resource, you really just want to have nothing change except for the specific value(s) passed in by the user. 

For example, if someone is updating the notes, you don't want languages to be reset to an empty list, do you? In that situation, the notes is the only thing that should change, nothing else should.

michizhou:
Thanks for sharing your thoughts! Going off of your concerns, I have completely refactored the code used to implement the updating of resources. Please let me know if other changes should be made.

aaron-suarez:
@michizhou After running it and playing with it, I realized that I led you in the wrong direction before. Sorry about that.

`request.args` is for query parameters that are sent as part of the query string. I didn't intend for it to be query params, I intended for it to be JSON form data. In order to get at the JSON, you need to use the method `request.get_json()`. 

```python
return set_resource(id, request.get_json(), db)
```

And then to actually set them

```python
def set_resource(id, json, db):
...

            if json.get('name'):
                resource.name = json.get('name')
            if json.get('url'):
                resource.url = json.get('url')
            if json.get('paid'):
                resource.paid = json.get('paid')
            if json.get('notes'):
                resource.notes = json.get('notes')
```

The category and languages are a bit trickier because they require you to get the Category and Languages if they exist and to create new ones if they do not exist. If you need help with the particular syntax for that (since you can't try it out yourself) I can do that tomorrow. I need to get going for the night.

If you're itching to get this merged, we can go ahead and just support updating the name, url, paid, and notes for now and open up a separate issue for updating the category and languages. This has been a bit of a long process, and I don't want to delay it any longer than we have to, I want to get your changes in.

michizhou:
@michizhou No worries! Mistakes like this happen. In any case, I am willing to wait until tomorrow so that you can help me with the syntax for updating `category` and `languages`. However, I also would like to get my changes merged, so if updating `category` and `languages` will take considerable time (i.e. at least half a day), I'd suggest opening a separate issue for updating those two fields.

michizhou:
@aaron-suarez I just pushed some changes I made to have JSON form data passed into `set_resource` in place of the query parameters mistakenly used before. I would appreciate it if you could help with the syntax for updating `category` and `languages` if you are available now.

michizhou:
@aaron-suarez That sounds great! I just added the lines for passing in `category` and `languages` to the scripts `set_resource` method. Also, do you happen to know why the build is failing now? I think it might have something to do with using `request.get_json()` or something.

aaron-suarez:
Ugh... Ok, so this works when creating a _new_ language or category, but it fails when trying to change it to an existing language or category. We need to pull in existing languages/categories from the DB to get those objects and only create a new one if it doesn't already exist. Code sample incoming

aaron-suarez:
```python
    try:
        resource = Resource.query.get(id)
        languages_list = Language.query.all()
        categories_list = Category.query.all()

        language_dict = {l.key(): l for l in languages_list}
        category_dict = {c.key(): c for c in categories_list}

    except MultipleResultsFound as e:
        print_tb(e.__traceback__)
        print(e)

    except NoResultFound as e:
        print_tb(e.__traceback__)
        print(e)

    finally:
        if resource:
            if json.get('languages'):
                langs = []
                for lang in json.get('languages') or []:
                    language = language_dict.get(lang)
                    if not language:
                        language = Language(name=lang)
                    langs.append(language)
                resource.languages = langs
            if json.get('category'):
                resource.category = category_dict.get(json.get('category'), Category(name=json.get('category')))
            if json.get('name'):
                resource.name = json.get('name')
            if json.get('url'):
                resource.url = json.get('url')
            if json.get('paid'):
                resource.paid = json.get('paid')
            if json.get('notes'):
                resource.notes = json.get('notes')
```

aaron-suarez:
We eventually need to do some error checking, this kind of just fails silently right now. But we'll handle that in a different PR. For now, this works, and that's good enough for me

michizhou:
How is the fix for the dependency management issue coming along? Is it close to b in merged with this PR?

aaron-suarez:
I was trying to get the code working for this PR, now I'm working on reviewing #47. I have to go to dinner shortly, so I might be able to merge that in first, but I want to make sure it works properly before I just merge it in. If not tonight, it will be tomorrow morning

michizhou:
@aaron-suarez I just managed to incorporate the modified implementation for updating `category` and `languages` in a resource. Besides the dependency management issue, is this PR ready to merge?

aaron-suarez:
@michizhou #51 fixes the travis failure. Please rebase to master.

aaron-suarez:
Thanks for all your work on this, @michizhou!

PR #39, opened 2018-10-20
Added an endpoint for creating a new resource

### LINKED ISSUE #33: Need POST /resources Route
This endpoint should create a new resource based on JSON form data sent with the POST.

*Required attributes*:
`category` - A string category. Endpoint should add it to existing category or create a new category if necessary
`languages` - An array of strings representing each language associated with this resource. Endpoint should add any new languages to the languages table and associate this resource with any existing languages that are specified
`name` - A string that's the name of the resource
`url` - A string that is a valid URL. (Ideally, the endpoint would validate this URL by performing a GET request and ensuring it receives a 200 response, but that's not really necessary for the MVP, just something to think about)

*Optional attributes*:
`paid` - Boolean. True if the resource requires payment to use. False if it's a free resource. Defaults to false
`notes` - String. Any notes that are relevant to the resource

michizhou:
Added code for a new endpoint that is designed to create a resource based on the JSON form data parameters passed in a POST request. Creates a new resource instance and initializes its attributes to the values of the passed parameters from the request. Closes #33.

### DIFF
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -36,9 +36,9 @@ def resource_set(id, params="", category=None, languages=[], name=None, url=None
     return set_resource(id, param_list)
 
 
-@bp.route('/resources/<int:id>/<string:category>/<string:languages>/<string:name>/
-          <string:url>/<string:params>', methods=['POST'])
-def create(id, params="", category, languages, name, url, paid=False, notes=None):
+@bp.route('/resources/<category>/<languages>/<name>/<string:url>/<string:params>', 
+          methods=['POST'])
+def create_new(params="", category, languages, name, url, paid=False, notes=None):
     new_resource = Resource()
     optional_params = params.split('/')
     if len(optional_params) > 0:
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -127,7 +127,7 @@ def get_languages():
         return jsonify(language_list)
 
 
-def get_attributes():
+def get_attributes(json):
     languages_list = Language.query.all()
     categories_list = Category.query.all()
 
@@ -160,9 +160,9 @@ def set_resource(id, json, db):
     finally:
         if resource:
             if json.get('languages'):
-                resource.languages = get_attributes()[0]
+                resource.languages = get_attributes(json)[0]
             if json.get('category'):
-                resource.category = get_attributes()[1]
+                resource.category = get_attributes(json)[1]
             if json.get('name'):
                 resource.name = json.get('name')
             if json.get('url'):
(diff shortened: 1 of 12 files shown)


michizhou:
@aaron-suarez Similar to my latest commit for #38, the build her also seems to be failing due to a similar non-existent "invalid syntax" error. I would appreciate it if you look into these. Thanks!

aaron-suarez:
Now that we've merged #38, I'll be happy to review this. Please fix the conflicts and I'll take a look

michizhou:
@aaron-suarez I just managed to update the code for creating a new resource. Please let me know if I need to make any changes in your subsequent review.

app/api/routes.py
```diff
@@ -174,3 +180,22 @@ def set_resource(id, json, db):
...
         else:
             return jsonify({})
+
+
+def create_resource(json, db):
+    new_resource = Resource(
+        name=json.get('name'),
+        url=json.get('url'),
+        category=get_attributes(json)[1],
+        languages=get_attributes(json)[0],
```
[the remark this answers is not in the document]
michizhou: I'll be sure to incorporate similar logic into the `set_resource` function since it seems doable.

aaron-suarez:
Great thanks for making these changes. I'm going to actually run the code and make sure it works before I merge it. I might not be online tonight, but if I don't get to it tonight, I'll get it tomorrow.

app/api/routes.py
```diff
@@ -17,12 +17,14 @@ def resources():
     return get_resources()
 
 
-@bp.route('/resources/<int:id>', methods=['GET', 'PUT'])
+@bp.route('/resources/<int:id>', methods=['GET', 'PUT', 'POST'])
```
aaron-suarez: The route needs to be `POST /resources` due to the fact that there is no ID for a brand new resource until the database assigns one.

Please move the `POST` method and the call to `create_resource` up to this method here https://github.com/OperationCode/resources_api/blob/63c67b898caf797eb8497c332a37a187cb05f2e8/app/api/routes.py#L15

app/api/routes.py
```diff
@@ -174,3 +181,20 @@ def set_resource(id, json, db):
...
+        name=json.get('name'),
+        url=json.get('url'),
+        category=categ,
+        languages=langs,
+        paid=json.get('paid'),
+        notes=json.get('notes'))
+
+    try:
+        db.session.add(new_resource)
+    except Exception as e:
```
aaron-suarez: The data isn't getting saved in the database, you need to add a call to `db.session.commit()` in order to save it. And when you add that, please also add a call to `db.session.rollback()` in the exception handler so that if anything goes wrong, the db stays in a consistent state. Should look like this:
```python
    try:
        db.session.add(new_resource)
        db.session.commit()
    except Exception as e:
        db.session.rollback()
        print('exception', e)
```

aaron-suarez:
Those should be the last two changes needed. After that, we can get this merged

michizhou:
@aaron-suarez Just pushed through your recommended changes. It should be ready to merge!

aaron-suarez:
Thanks @michizhou!

Issue #29, opened 2018-10-18
GET /resources Route should take query params for language and category


aaron-suarez:
The original vision of routes was flawed. We originally wanted to have routes called `GET /categories/{category}` and `GET /languages/{language}` to get all of the resources for a category and language. After some discussion, @wimo7083 and I decided that it makes much more sense to have the single route for `GET /resources` take optional query parameters specifying the category and language. So a valid request to this endpoint can take `page`, `page_size` (these two are addressed in #21) as well as `language` and `category`. So the request might look something like this:

`GET /resources?category=books&language=ruby&page=2&page_size=5`

This issue will be considered fixed when the query params for `language` and `category` are accepted by the endpoint and return the filtered results.

heavysaturn:
agreed. I will update #28 to do this instead.

aaron-suarez:
This was fixed in #42

PR #42, opened 2018-10-22
Params for filtering resources on language or category. 

heavysaturn:
Opening a new pull request because the other one fell victim to the GitHub problems that have been happening today and yesterday.

This PR solves #29, allowing you to filter GET requests to the resources endpoint by language, category, or both. The requests are still paginated in the same way, with the filtering being applied _before_ the pagination is done. All filters are caps insensitive.

### DIFF
--- a/app/api/__init__.py
+++ b/app/api/__init__.py
@@ -2,3 +2,6 @@ from flask import Blueprint
 
 bp = Blueprint('api', __name__)
 
+# We need to import the routes here so they will
+# bind to the blueprint before the blueprint is registered.
+from app.api import routes  # noqa
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -10,6 +10,7 @@ from app.models import Language, Resource, Category
 from app import Config
 from app.utils import Paginator
 
+
 # Routes
 @bp.route('/resources', methods=['GET'])
 def resources():
@@ -97,7 +98,9 @@ def get_resources():
         else:
             query = Resource.query
 
-        resource_list = [resource.serialize for resource in resource_paginator.items(query)]
+        resource_list = [
+            resource.serialize for resource in resource_paginator.items(query)
+        ]
 
     except Exception as e:
         print_tb(e.__traceback__)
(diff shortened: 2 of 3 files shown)


aaron-suarez:
There's some things going on here I don't quite understand. We may need to optimize the queries at some point, and perhaps some of this will be solved by some of the data cleaning issues. But I pulled it down and tried it out and everything worked as I expected it. I like the change to the paginator, less parameters is simpler.

Thanks for this @heavysaturn

heavysaturn:
@aaron-suarez I'd be happy to shed some light on any particulars you don't understand, or add more comments to clarify parts that need clarification, if necessary.

As for the queries, because of the way SQLA and Postgres work, these are already quite fast. SQL Alchemy will turn these kinds of queries into highly performant single raw SQL queries, something like
```python
query = Resource.query.filter(
    Resource.languages.any(
        Language.name.ilike(language)
    )
)
```

will be turned into something like this:
```sql
SELECT * 
FROM resource 
WHERE %s 
ILIKE ANY(resource.languages)
```

and then `language` will be interpolated into the `%s`. 

This is a case insensitive group membership comparison that checks if the language is in the list of languages inside the ANY. It's basically as fast as it gets. Everything is handled in a single query, and everything is handled by the database rather than the backend.

aaron-suarez:
I am not familiar with `ILIKE`, but I know that `LIKE` is slower than `=` and it can't take advantage of indexing if I understand correctly. However, in this particular app, we're letting people search with arbitrary strings, so perhaps that's faster than string comparisons in python...

heavysaturn:
@aaron-suarez ILIKE is just case insensitive LIKE. I did this because I noticed some of the resources had differences in case - like `python` and `Python`. This way, the filter doesn't care about case at all and finds all python resources like it should.

It's true that `LIKE` is slower than `=` and it's possible that you could shave off some microseconds by optimizing these queries, but considering how few rows are ever going to be in this database, this would fall under what Donald Knuth refers to as "premature optimization" (which he then goes on to characterize as "the root of all evil"). These queries are already so fast that optimizing them is, at best, a waste of time, and, at worst, going to hurt readability.

Issue #13, opened 2018-10-07
No setup instructions for Windows


aaron-suarez:
The README explains the development setup steps for Mac, but not for Windows. Eventually, we want to move to some type of Docker setup, but for the moment, it would be helpful to describe the steps to get this working for contributors working on Windows machines.

heavysaturn:
I could probably add this, but yeah using docker and docker-compose would reduce the setup to:

1. install docker and docker-compose
2. fork this repo
3. clone your fork
4. run `docker-compose up` in the clone folder
5. if you're on windows or mac, run `docker-machine ip` to figure out what your docker ip is. if you're on *nix, you can just connect to localhost instead.
6. server is now running! go to `http://YOUR_DOCKER_IP:5000/api/v1/` to test.

steps 5 and 6 could also be displayed when you run `docker-compose up` if we're a bit clever about it. we could detect the OS, run the command to get the ip if necessary, and build a print that tells the user where exactly the server is running, just like flask normally does when you use `flask run`

heavysaturn:
it'd be _even better_ if we managed to get this to play well with pipenv as well, using docker-compose only for the postgres database and pipenv to build the venv, install the packages from pip, export the environment variables, and expose simple scripts the user could run.

- `pipenv run init` to set up the database with the initial data and make everything ready to run. here, the environment variables could be set up in such a way that windows and mac users would connect to a remote postgres located on the docker ip.
- `pipenv run start` to run the flask app,
- `pipenv run lint` to lint the app
- `pipenv run test` to run all the unit tests and show results
- `pipenv run hook` to create a pre-commit hook that lints and runs all tests before allowing a commit.

aaron-suarez:
We can do a few of those with click commands though, right? We already have some of that for the migrate/upgrade for example. Does pipenv offer some special benefit?

heavysaturn:
basically pipenv is a marriage of requirements.txt, pip and venv. You make a `Pipfile` that says which packages to use, which version of python to use for the venv, and as a bonus feature you can define these run scripts.

The user never has to create a venv, activate the venv, install packages (we can remove `requirements.txt` completely), or anything else. they just sync the pipenv, and then use the scripts to interact with stuff inside it. if they need access to the venv shell, there's a `pipenv shell` command for that too

heavysaturn:
[...]

PR #43, opened 2018-10-22
add windows setup README.md

VeraButler:
This needs to be tested @aaron-suarez. Long_common_name offered to help on this. 

This includes Chocolatey install and postgres setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,22 @@ First, you'll need to configure a database. Once this project is deployed, we'll
 
 ### Windows setup
 
-Coming Soon!
+## Chocolatey and Postgres (Needs Testing)
+1. [Install Choclatey](https://chocolatey.org/docs/installation#installing-chocolatey) if you do not already have it installed.
+2. Add `C:\ProgramData\chocolatey\bin` to your paths (type path in search, click Edit environment varaibles for you account).
+3. In your cmd.exe (admin) terminal run `choco install postgresql --params '/Password:<assigned to postgres user>` (cmd.exe (admin) can be found by pressing `win + x`).
+4. Upgrade postgresql: `choco upgrade postgresql`.
+5. Add `C:\Program Files\PostgreSQL\10\bin` to your paths.
+6. Start postgres: `psql -U postgres` you will see postgres-# in the terminal.
+7. Create your user with `CREATE USER name;` The terminal will print CREATE ROLE.
+8. Alter your role with permission to create a database with `ALTER USER name WITH CREATEDB CREATEROLE;` The terminal will print ALTER ROLE
+9. Check everything is in order with `\du`. The terminal will print a table with Role name, Atrributes, and Member of columns. If your user is listed you are good to go.
+10. Quit with `\q` or `ctr + c`.
+11. Start psql with new username with `psql`. You should see `user=>`.
+12. Create a database with: CREATE DATABASE resources OWNER user;
+
+<--NEED TO ADD THE REST-->
+
 
 ### Mac setup
 


aaron-suarez:
Ok, looking good so far. If I get some time before @wimo7083 does, I'll fire up a Windows VM and see if it works for me. Thanks for this! Fixes #13

README.md
```diff
@@ -10,7 +10,22 @@ First, you'll need to configure a database. Once this project is deployed, we'll
 
 ### Windows setup
 
-Coming Soon!
+## Chocolatey and Postgres (Needs Testing)
+1. [Install Choclatey](https://chocolatey.org/docs/installation#installing-chocolatey) if you do not already have it installed.
+2. Add `C:\ProgramData\chocolatey\bin` to your paths (type path in search, click Edit environment varaibles for you account).
+3. In your cmd.exe (admin) terminal run `choco install postgresql --params '/Password:<assigned to postgres user>` (cmd.exe (admin) can be found by pressing `win + x`).
```
wimo7083: https://chocolatey.org/packages/postgresql

apparently choco is down, so I'll have to check later.

wimo7083: I think this is missing a closing single quote.

wimo7083: If we are able to explicitly set a username, I'd prefer that params. i'll look at the documentation when it's back up. 
I'd also like to include a full example of the command that requires no user guessing what the `<assigned to postgres user>` means.

README.md
```diff
@@ -10,7 +10,22 @@ First, you'll need to configure a database. Once this project is deployed, we'll
...
+4. Upgrade postgresql: `choco upgrade postgresql`.
+5. Add `C:\Program Files\PostgreSQL\10\bin` to your paths.
+6. Start postgres: `psql -U postgres` you will see postgres-# in the terminal.
```
wimo7083: Add step telling them to restart terminal so path updates.

README.md
```diff
@@ -10,7 +10,22 @@ First, you'll need to configure a database. Once this project is deployed, we'll
...
+7. Create your user with `CREATE USER name;` The terminal will print CREATE ROLE.
+8. Alter your role with permission to create a database with `ALTER USER name WITH CREATEDB CREATEROLE;` The terminal will print ALTER ROLE
+9. Check everything is in order with `\du`. The terminal will print a table with Role name, Atrributes, and Member of columns. If your user is listed you are good to go.
+10. Quit with `\q` or `ctr + c`.
+11. Start psql with new username with `psql`. You should see `user=>`.
+12. Create a database with: CREATE DATABASE resources OWNER user;
```
wimo7083: Please include backticks for this command. 
`CREATE DATABASE resources OWNER user;`

README.md
```diff
@@ -10,7 +10,22 @@ First, you'll need to configure a database. Once this project is deployed, we'll
 
 ### Windows setup
 
-Coming Soon!
+## Chocolatey and Postgres (Needs Testing)
+1. [Install Choclatey](https://chocolatey.org/docs/installation#installing-chocolatey) if you do not already have it installed.
+2. Add `C:\ProgramData\chocolatey\bin` to your paths (type path in search, click Edit environment varaibles for you account).
```
wimo7083: I'd maybe add a whole group related to adding values to the path. Adding to path is one of the most common issues I see with new people and python projects (and windows). Even something like a link to a url would be good.

wimo7083:
Looks like there is only one param and this is the format (now that site is up)
image

README.md
aaron-suarez: On my windows 10 vm using admin powershell (I know that wasn't the instruction, but whatever), I ran `choco install postgresql --params '/Password:getmein'` and it still prompted me for a password for the postgres user. Not sure what happened there

README.md
```diff
@@ -10,7 +10,22 @@ First, you'll need to configure a database. Once this project is deployed, we'll
...
+3. In your cmd.exe (admin) terminal run `choco install postgresql --params '/Password:<assigned to postgres user>` (cmd.exe (admin) can be found by pressing `win + x`).
+4. Upgrade postgresql: `choco upgrade postgresql`.
+5. Add `C:\Program Files\PostgreSQL\10\bin` to your paths.
+6. Start postgres: `psql -U postgres` you will see postgres-# in the terminal.
+7. Create your user with `CREATE USER name;` The terminal will print CREATE ROLE.
+8. Alter your role with permission to create a database with `ALTER USER name WITH CREATEDB CREATEROLE;` The terminal will print ALTER ROLE
+9. Check everything is in order with `\du`. The terminal will print a table with Role name, Atrributes, and Member of columns. If your user is listed you are good to go.
+10. Quit with `\q` or `ctr + c`.
+11. Start psql with new username with `psql`. You should see `user=>`.
```
aaron-suarez: I got `psql: FATAL: password authentication failed for user "aaron"` Your instructions didn't say how to set a password. If I try to press "enter" without giving a password I get `psql: fe_sendauth: no password supplied`

README.md
VeraButler: Ok.  That is from the Chocolatey docs and params password is assigned to pastgres before you create a user.  Maybe better wording would be: <choose a password>. Your password will be assigned to the default postgres superuser.

VeraButler: The hiccups I ran into were with the password.  Maybe it would be better to just let the terminal prompt the user. Thoughts?

README.md
VeraButler: Yes I had many issues with the password.  It seems it did not set when prompted by the terminal.  At first I thought maybe it was a typo, but now I don't know.

README.md
VeraButler: Also, it should be powershell because that is what is listed in the win + x menu.

README.md
VeraButler: "For anyone else hitting an issue where the password is not being read from the cli parameter, it turned out to be caused by the package alias for me - calling "choco install postgresql" failed, but "choco install postgresql10" works."

Found this in the comments on the docs https://chocolatey.org/packages/postgresql.

VeraButler: Also, this is how I updated my password. https://stackoverflow.com/questions/18587710/change-reset-postgresql-user-password-on-windows-7

README.md
VeraButler: Added

README.md
VeraButler: I removed the --param all together after reading what Aaron said about it prompting for a password anyways.

README.md
VeraButler: Added

VeraButler:
This may be a better link for resetting your password. https://vialect.net/blog/reset-the-password-for-postgresql-on-windows/

README.md
```diff
@@ -10,7 +10,23 @@ First, you'll need to configure a database. Once this project is deployed, we'll
...
 ### Windows setup
 
-Coming Soon!
+## Chocolatey and Postgres (Needs Testing)
+1. [Install Chocolatey](https://chocolatey.org/docs/installation#installing-chocolatey) if you do not already have it installed.
+2. Add `C:\ProgramData\chocolatey\bin` to your paths (type path in search, click Edit environment varaibles for you account).
+3. In your powershell (admin), run `choco install postgresql10` (powershell (admin) can be found by pressing `win + x`).
+4. Upgrade postgresql: `choco upgrade postgresql`.
+5. Add `C:\Program Files\PostgreSQL\10\bin` to your [paths](https://helpdeskgeek.com/windows-10/add-windows-path-environment-variable/).
+6. Restart your admin terminal to update the path.
```
aaron-suarez: I don't like having them restart the terminal because of the path, and I also don't like having to open up the window where environment variables are stored twice.

Can we just have them add both things to the path in one step before the powershell session is ever started, then it's already correct once they actually install it? (Not sure if Windows would complain about this, I know this would be fine on Linux)

VeraButler: I thought about that. I do not /think/ it would cause complaints. I left it separate for two reasons: 1. unsure about the complaints, 2. I added to Development Notes to see step 22 for the FLASK_APP=run.py.

README.md
```diff
@@ -10,7 +10,46 @@ First, you'll need to configure a database. Once this project is deployed, we'll
 
 ### Windows setup
 
-Coming Soon!
+## Chocolatey and Postgres: (Needs Testing)
```
aaron-suarez: Can you please make these headings four `#`? They should be smaller than `Windows setup`

README.md
```diff
@@ -10,7 +10,46 @@ First, you'll need to configure a database. Once this project is deployed, we'll
...
+7. Start postgres: `psql -U postgres` you will see postgres-# in the terminal.
+8. Create your user with `CREATE USER name;` The terminal will print CREATE ROLE.
+9. Alter your role with permission to create a database with `ALTER USER name WITH CREATEDB CREATEROLE;` The terminal will print ALTER ROLE
+10. Check everything is in order with `\du`. The terminal will print a table with Role name, Atrributes, and Member of columns. If your user is listed you are good to go.
+11. Quit with `\q` or `ctr + c`.
+12. Start psql with new username with `psql`. You should see `name=>`.
+13. Create a database with: `CREATE DATABASE resources OWNER name;`.
+14. Connect to the resources database: `\c resources`. If you see resources => you are ready to move on to the next steps.
+## Set Up and Activate your Environment:
+15. Set the connection information in a [user environment variable](https://docs.microsoft.com/en-us/windows/desktop/shell/user-environment-variables):
```
aaron-suarez: I think we can add this step to "Step 2" and add all of the variables all in one place, huh?

README.md
```diff
@@ -10,7 +10,46 @@ First, you'll need to configure a database. Once this project is deployed, we'll
...
+- Variable name: `FLASK_APP`
+- Variable Value: `run.py`
+- Optionally, enable debugging by setting the environment to development with the Variable Value: `development`
+24. Restart powershell.
+23. Create the tables in your database with `flask db_migrate create_tables`
+24. Tell flask that your database is up to date with `flask db stamp head`
+25. Populate your database with the resources `flask db_migrate init`
+26. Start your development server with `flask run` and you're ready to go!
+
+27. Check your work: `Ctrl + c` then `psql resources`. The terminal pointer will say `resources=>`. Then `\dt` to view the schema.
```
aaron-suarez: I think perhaps the check should be to navigate to `localhost:5000/api/v1/resources` and see a bunch of resources, huh? If you don't, then you know _something_ went wrong. Not everyone will know what they're looking for when they see the schema, and it doesn't prove that everything made it into those tables just because the schema is correct.

README.md
VeraButler: fixed

README.md
VeraButler: fixed

README.md
```diff
@@ -10,7 +10,45 @@ First, you'll need to configure a database. Once this project is deployed, we'll
...
+8. Create your user with `CREATE USER name;` The terminal will print CREATE ROLE.
+9. Alter your role with permission to create a database with `ALTER USER name WITH CREATEDB CREATEROLE;` The terminal will print ALTER ROLE
+10. Check everything is in order with `\du`. The terminal will print a table with Role name, Atrributes, and Member columns. If your user is listed you are good to go.
+11. Quit with `\q` or `ctr + c`.
+12. Start psql with new username with `psql`. You should see `name=>`.
+13. Create a database with: `CREATE DATABASE resources OWNER name;`.
+14. Connect to the resources database: `\c resources`. If you see `resources =>` you are ready to move on to the next steps.
+15. In powershell navigate to where you would like to clone a fork of your repository. `cd \path\to\directory`
+16. Clone your fork of the project `git clone {project_url} resources`
+17. Navigate to the clone directory `cd resources-api`
```
aaron-suarez: If you do `git clone {project_url} resources`, then it will create the project in a directory called `resources`, not `resources-api`

VeraButler: On my machine it cloned as resources_api. It is still a typo.

README.md
```diff
@@ -43,6 +81,7 @@ export SQLALCHEMY_DATABASE_URI=postgresql://aaron:password@127.0.0.1:5432/resour
 If you make changes to the models.py or other schemas, you need to run a migration and upgrade again:
 
 1. Set the FLASK_APP environment variable `export FLASK_APP=run.py` or `ENV:FLASK_APP = "run.py"`
+- Windows users follow Step 15 in the Windows setup.
```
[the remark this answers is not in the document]
VeraButler: fixed

aaron-suarez:
Ok, we're looking good. I'm going to run through the steps from front to back in a VM one more time to make sure there are no other snags. In the meantime, can you remove the commits to the routes.py and configs.py? That will fix the travis error.

VeraButler:
Do you have a guide for deleting a commit?

aaron-suarez:
I would do an interactive rebase. https://stackoverflow.com/questions/11113322/remove-4-commits-from-my-git-history?answertab=votes#tab-top

aaron-suarez:
If you run into trouble, ping me on slack and I'll walk you through it

VeraButler:
So I rebased to the commit before the Categories...bu there is still a build issue.  When I read what the issue is it is saying there is a problem with one of the dependencies "itsdangerous==1.0.0 ".

PR #46, opened 2018-10-24
Makefile changes

sethbergman:
I've added a couple dependencies and modified the Makefile. Let me know if this helps.


Dockerfile
```diff
@@ -12,7 +12,9 @@ COPY Pipfile* /app/
 RUN apt-get update \
     && apt-get install -y libpq-dev gcc \
     && rm -rf /var/lib/apt/lists/* \
-    && pip install --upgrade pip
+    && pip install --upgrade pip \
+    && pip install python-dotenv \
+    && pip install flask
```
wimo7083: Could you explain the purpose of installing flask with pip and pipenv? Is there something I'm missing?

Dockerfile
```diff
@@ -23,4 +25,4 @@ COPY app /app
 
 EXPOSE 8000
 
-CMD ["python", "-m", "flask", "run"]
+ENTRYPOINT ["python", "run.py"]
```
wimo7083: I tried to use ENTRYPOINT but  if you need to override it's better to use CMD.

Makefile
```diff
@@ -42,24 +43,23 @@ routes:
...
-	${DOCKER_COMPOSE} build
+	${DOCKER_COMPOSE} build --pull
 
 .PHONY: setup
 setup: build
-	${DOCKER_COMPOSE} run ${RESOURCES_CONTAINER}  flask db-migrate create-tables
-	${DOCKER_COMPOSE} run ${RESOURCES_CONTAINER}  flask db stamp head
-	${DOCKER_COMPOSE} run ${RESOURCES_CONTAINER}  flask db-migrate init
-
+	${DOCKER_COMPOSE} run ${RESOURCES_CONTAINER} -m flask db-migrate create-tables
```
wimo7083: For some reason I think that this doesn't work since it will call
`docker-compose run resources01 -m flask db stamp head`
This is missing `python` or needs to remove the `-m`?

Dockerfile
sethbergman: I'm running this on Linux Mint 19 with docker and docker-compose already installed. When I ran the `make`, `make build` and `make setup` commands my bash interpreter recommended installing those packages. Once I added them to the Dockerfile the shell had no more complaints.

Dockerfile
sethbergman: I also tried both and only got a graceful exit with `ENTRYPOINT`. I'm not sure why; it could have been something else, but I'm about 75% sure that reverting back to using `ENTRYPOINT` seemed to help.

Makefile
sethbergman: There are two spaces after `${RESOURCES_CONTAINER}` in the previous commit.

Dockerfile
sethbergman: It coud be that `CMD` needs `"run.py"` instead of `"run"`, but I'm not sure.

Makefile
```diff
@@ -1,6 +1,7 @@
 DOCKER := docker
 DOCKER_COMPOSE := docker-compose
 RESOURCES_CONTAINER := resources
+FLASK := flask
```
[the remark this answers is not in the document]
sethbergman: Done :rocket:

Dockerfile
sethbergman:
```sh
Step 12/12 : CMD ["python", "flask", "-m", "run.py"]
 ---> Running in a994d6d7a460
Removing intermediate container a994d6d7a460
 ---> 541ac4b41f7f

Successfully built 541ac4b41f7f
Successfully tagged resources:latest
Starting pg01 ... done
Recreating resources01 ... done
Attaching to pg01, resources01
pg01         | 2018-10-25 03:56:28.006 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
pg01         | 2018-10-25 03:56:28.006 UTC [1] LOG:  listening on IPv6 address "::", port 5432
pg01         | 2018-10-25 03:56:28.010 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
pg01         | 2018-10-25 03:56:28.030 UTC [19] LOG:  database system was shut down at 2018-10-25 03:56:12 UTC
pg01         | 2018-10-25 03:56:28.035 UTC [1] LOG:  database system is ready to accept connections
resources01  | (null): can't open file 'flask': [Errno 2] No such file or directory
resources01 exited with code 2
```

sethbergman:
```sh
Step 12/12 : CMD ["python", "-m", "flask", "run"]
 ---> Running in 999fb985a7f6
Removing intermediate container 999fb985a7f6
 ---> 799def31e222

Successfully built 799def31e222
Successfully tagged resources:latest
Starting pg01 ... done
Recreating resources01 ... done
Attaching to pg01, resources01
pg01         | 2018-10-25 03:58:30.034 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
pg01         | 2018-10-25 03:58:30.034 UTC [1] LOG:  listening on IPv6 address "::", port 5432
pg01         | 2018-10-25 03:58:30.039 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
pg01         | 2018-10-25 03:58:30.057 UTC [20] LOG:  database system was shut down at 2018-10-25 03:58:17 UTC
pg01         | 2018-10-25 03:58:30.062 UTC [1] LOG:  database system is ready to accept connections
resources01  |  * Serving Flask app "run.py"
resources01  |  * Environment: production
resources01  |    WARNING: Do not use the development server in a production environment.
resources01  |    Use a production WSGI server instead.
resources01  |  * Debug mode: off
resources01  |  * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)
```
The command `make` doesn't exit. The process stays open.

sethbergman:
```sh
Step 12/12 : ENTRYPOINT ["python", "run.py"]
 ---> Using cache
 ---> e56614d628d2

Successfully built e56614d628d2
Successfully tagged resources:latest
Creating pg01 ... done
Creating resources01 ... done
Attaching to pg01, resources01
pg01         | 2018-10-25 04:03:03.287 UTC [1] LOG:  listening on IPv4 address "0.0.0.0", port 5432
pg01         | 2018-10-25 04:03:03.287 UTC [1] LOG:  listening on IPv6 address "::", port 5432
pg01         | 2018-10-25 04:03:03.294 UTC [1] LOG:  listening on Unix socket "/var/run/postgresql/.s.PGSQL.5432"
pg01         | 2018-10-25 04:03:03.317 UTC [20] LOG:  database system was shut down at 2018-10-25 04:02:45 UTC
pg01         | 2018-10-25 04:03:03.321 UTC [1] LOG:  database system is ready to accept connections
resources01 exited with code 0
```
With this you get a exit code of 0 so that the other build steps can proceed. Correct me if I'm wrong about this. Shall we proceed with ` ENTRYPOINT ["python", "run.py"]`?

sethbergman: I've resoleved the conflicts in `Dockerfile` to use `ENTRYPOINT ["python", "run.py"]`. Let me know if this works for you.

wimo7083: First example:
```Step 12/12 : CMD ["python", "flask", "-m", "run.py"]```
this is definitely not right as the correct order is `python -m $module $module args`

Second example:
```Step 12/12 : CMD ["python", "-m", "flask", "run"]```
this gives the desired behavior as we would sometimes want to run it not in a daemon but want to view the logs as it occurs. 

third example:
``` ENTRYPOINT ["python", "run.py"]```
I don't know the use of a simple `make run`, `make` or `make all` command  that just verifies it can start and then doesn't. 

apparently if one is defined at runtime `CMD` or `ENTRYPOINT` they have the same effect. 

override `ENTRYPOINT`
`docker run --entrypoint [my_entrypoint] test  `
override `CMD`
`docker run test [command 1] [arg1] [arg2]  `

discussion of the [two](http://www.johnzaccone.io/entrypoint-vs-cmd-back-to-basics/) 
```
Use ENTRYPOINT if you don't want developers to change the executable that is run when the container starts. You can think of your container as an "executable wrapper". A good strategy is to define a "stable" combination of executable + parameters as the ENTRYPOINT. Then you can (optionally) specify a default CMD that developers can easily override.
```
it looks like if you do a `CMD` and `ENTRYPOINT`

it has the effect of concatenating the two 
`docker run $container`
`$CMD ENTRYPOINT`

sethbergman: So would you prefer defining both `ENTRYPOINT` + `CMD` in the Dockerfile? Also, in `.travis.yml` we may need to change `- sleep 30` to `- sleep 10` as the Travis builds have a timeout of 10 minutes anyhow. If we leave a process open I'm pretty sure that the build will just timeout after running the flask command. We should look at using the matrix build methods if we're going to use the flask command IMO.

sethbergman: Nevermind on the `-sleep 30` part. I just examined the logs and it's 30 seconds, not minutes. :zzz:

sethbergman:
@wimo7083 can you check this and provide feedback when you get a chance? Thanks!

Dockerfile
sethbergman: I've added both `ENTRYPOINT ["python", "run.py"]` and `CMD ["python", "-m", "flask", "run"]` in Dockerfile.

PR #47, opened 2018-10-26
add windows setup steps

VeraButler:
this should fix the travis problem

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ This project provides an API for storing and retrieving learning resources that 
 
 First, you'll need to configure a database. Once this project is deployed, we'll be using PostgreSQL. To set up a psql instance locally, follow the instructions for your OS:
 
-### [Windows setup](#windowsSetup)
+### Windows Setup
 
 1. [Install Chocolatey](https://chocolatey.org/docs/installation#installing-chocolatey) if you do not already have it installed.
 2. Add Chocolately and postgresql to your paths.   
@@ -81,7 +81,7 @@ export SQLALCHEMY_DATABASE_URI=postgresql://aaron:password@127.0.0.1:5432/resour
 If you make changes to the models.py or other schemas, you need to run a migration and upgrade again:
 
 1. Set the FLASK_APP environment variable `export FLASK_APP=run.py` or `ENV:FLASK_APP = "run.py"`
-- Windows users see steps 2 & 3 under [Windows Setup](#windowsSetup)
+- Windows users see steps 2 & 3 under Windows Setup
 2. Run the migration `flask db migrate`
 3. Upgrade to the latest migration `flask db upgrade`
 
--- a/README.md
+++ b/README.md
@@ -8,9 +8,47 @@ This project provides an API for storing and retrieving learning resources that 
 
 First, you'll need to configure a database. Once this project is deployed, we'll be using PostgreSQL. To set up a psql instance locally, follow the instructions for your OS:
 
-### Windows setup
+### [Windows setup](#windowsSetup)
 
-Coming Soon!
+1. [Install Chocolatey](https://chocolatey.org/docs/installation#installing-chocolatey) if you do not already have it installed.
+2. Add Chocolately and postgresql to your paths.   
+- Search --> Edit the [system environment variables](https://docs.microsoft.com/en-us/windows/desktop/shell/user-environment-variables) --> Environment Variables... --> Path --> Edit... --> New --> Enter `C:\ProgramData\chocolatey\bin` --> OK
+- New --> `C:\Program Files\PostgreSQL\10\bin` --> OK
+3. Set the System Variables.
+- Click the bottom New... -->
+- Connection Variable:
+    - Variable Name: `SQLALCHEMY_DATABASE_URI`
+    - Variable Value: `postgresql://aaron:password@127.0.0.1:5432/resources` (instead of aaron and password put what user name and password you want to use for postgresql) 
+- FLASK_APP Variable:
+    - Variable name: `FLASK_APP`
+    - Variable Value: `run.py`
+    - Optionally, enable debugging by setting the environment to development with the Variable Value: `development`
+4. Close out of Environment Variables and System Properites.
+5. In your powershell (admin), run `choco install postgresql10`. Follow the prompts.
+- powershell (admin) can be found by pressing `win + x`. 
+- Make sure you use the admin powershell throughout all the steps.
+6. Upgrade postgresql: `choco upgrade postgresql`.
+7. Start postgres: `psql -U postgres` you will see postgres-# in the terminal.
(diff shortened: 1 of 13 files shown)


aaron-suarez:
Ok, this is looking good. I'm going to try to run through it either tonight or tomorrow morning (depending on how much time I have tonight) and if I can go through all the steps in a VM and nothing gives me an error, I'll merge this in.

Thanks for your effort here!

README.md
```diff
@@ -8,9 +8,47 @@ This project provides an API for storing and retrieving learning resources that
...
+5. In your powershell (admin), run `choco install postgresql10`. Follow the prompts.
+- powershell (admin) can be found by pressing `win + x`. 
+- Make sure you use the admin powershell throughout all the steps.
+6. Upgrade postgresql: `choco upgrade postgresql`.
+7. Start postgres: `psql -U postgres` you will see postgres-# in the terminal.
+8. Create your user with `CREATE USER name;` The terminal will print CREATE ROLE.
+9. Alter your role with permission to create a database with `ALTER USER name WITH CREATEDB CREATEROLE;` The terminal will print ALTER ROLE
+10. Check everything is in order with `\du`. The terminal will print a table with Role name, Atrributes, and Member columns. If your user is listed you are good to go.
+11. Quit with `\q` or `ctr + c`.
+12. Start psql with new username with `psql`. You should see `name=>`.
```
aaron-suarez: We need a command that one can copy and paste. I assumed the command was `psql -U aaron` but that didn't work for me. There needs to be no ambiguity here.

README.md
```diff
@@ -8,9 +8,47 @@ This project provides an API for storing and retrieving learning resources that
...
+13. Create a database with: `CREATE DATABASE resources OWNER name;`.
+14. Connect to the resources database: `\c resources`. If you see `resources =>` you are ready to move on to the next steps.
+15. In powershell navigate to where you would like to clone a fork of your repository. `cd \path\to\directory`
+16. Clone your fork of the project `git clone {project_url} resources`
```
aaron-suarez: This fails due to git not being installed. Need link to git install instructions.

README.md
```diff
@@ -8,9 +8,47 @@ This project provides an API for storing and retrieving learning resources that
...
+17. Navigate to the clone directory `cd resources`
+18. [Create a virtual environment](https://docs.python.org/3/library/venv.html) called venv `python -m virtualenv venv`.
```
aaron-suarez: Similarly, this fails due to python not being installed. Need to link python install instructions.

It might be a good idea to (at the very top) list everything that's needed and install instructions for them, that way if people already have them, they can skip past it but if they don't, they can set up their environment properly before they run commands that will fail.

aaron-suarez: Once you do install python, you get the error "No module named virtualenv". Need to specify to pip install that

README.md
```diff
@@ -8,9 +8,47 @@ This project provides an API for storing and retrieving learning resources that
...
+19. Set Execution Policy to unrestricted `Set-ExecutionPolicy Unrestricted -Force`
+20. Activate virtual environment `venv\Scripts\Activate.ps1`
+21. Install the required dependencies `pip install -r requirements.txt`
+22. Create the tables in your database with `flask db_migrate create_tables`
```
aaron-suarez: I'm not sure why or how, but I had to run `flask db-migrate create-tables` rather than `flask db_migrate create_tables`... If that's a windows thing, this command needs to change, but if it could be either/or, we should call out that that's a possibility and that running `flask --help` will list the available commands

README.md
aaron-suarez: Also, I didn't even need to quit here. I was able to do `CREATE DATABASE resources OWNER aaron;` as the postgres user and that worked just fine. If we don't need the extra step of logging out of postgres and logging back in as the user, we should just take out steps 11 and 12 altogether.

Did you need to do this to get it to work?

README.md
```diff
@@ -6,11 +6,61 @@ This project provides an API for storing and retrieving learning resources that
...
+11. Check everything is in order with `\du`. The terminal will print a table with Role name, Atrributes, and Member columns. If your user is listed you are good to go.
+12. Create a database with: `CREATE DATABASE resources OWNER name;`.
+13. Connect to the resources database: `\c resources`. If you see `resources =>` you are ready to move on to the next steps.
+14. Navigate to the cloned repo directory `cd \path\to\resources`. (see step 6 under getting started)
+15. Install virtualenv if you do not have it. `pip install virtualenv`.
+16. [Create a virtual environment](https://docs.python.org/3/library/venv.html) called venv `python -m virtualenv venv`.
+17. Set Execution Policy to unrestricted `Set-ExecutionPolicy Unrestricted -Force`
+18. Activate the virtual environment `venv\Scripts\Activate.ps1`
+19. Install the required dependencies `pip install pipenv`
+20. Create the tables in your database with `flask db-migrate create-tables`
```
aaron-suarez: Need a call to `pipenv install --dev` after `pip install pipenv`

README.md
```diff
@@ -25,24 +75,26 @@ Note: if you try to use the same username as the user you are logged in as, you
...
-8. Clone your fork of the project `git clone {project_url} resources`
-9. Change directory into the project `cd resources`
-10. Create a virtual environment called venv `python -m virtualenv venv`
+8. Navigate to the cloned repo directory `cd \path\to\resources`. (see step 6 under getting started)
+9. Install virtualenv if you do not have it. `pip install virtualenv`.
+10. [Create a virtual environment](https://docs.python.org/3/library/venv.html) called venv `python -m virtualenv venv`.
 11. Activate virtual environment `source venv/bin/activate`
-12. Install the required dependencies `pip install -r requirements.txt`
+12. Install the required dependencies `pip install pipenv`
 13. Set the FLASK_APP environment variable `export FLASK_APP=run.py` or `ENV:FLASK_APP = "run.py"`
```
aaron-suarez: Also add the call to `pipenv install --dev` here as well

README.md
```diff
@@ -6,11 +6,61 @@ This project provides an API for storing and retrieving learning resources that
...
+    - Variable Value: `run.py`
+    - Optionally, enable debugging by setting the environment to development with the Variable Value: `development`
+3. Close out of Environment Variables and System Properites.
+4. Start your administrative shell. `ctrl + x` --> Windows Powershell (Admin).
+- **Make sure you use the _admin powershell_ throughout all the steps.**
+5. [Install Chocolatey](https://chocolatey.org/docs/installation#installing-chocolatey) if you do not already have it installed.
+6. In your powershell (admin), run `choco install postgresql10`. Follow the prompts.
+7. Upgrade postgresql: `choco upgrade postgresql`.
+8. Start postgres: `psql -U postgres` you will see postgres-# in the terminal.
+9. Create your user with `CREATE USER name;` The terminal will print CREATE ROLE.
```
aaron-suarez: Need to add a password here. The command should be `CREATE USER name WITH PASSWORD 'password';`

PR #49, opened 2018-10-26
Update README.md

EndermanOfCoding:


aaron-suarez:
I like it as a comma, not a semi colon. If you'd like to contribute to the project, please join us on slack and we can give you an issue. https://operationcode.org/join

PR #51, opened 2018-10-27
Fix the dependency issues causing travis failure

aaron-suarez:
Travis was failing for every PR based on a dependency issue. This fixes that problem.


aaron-suarez:
Also, credit for this should really go to @VeraButler because she fixed it on her branch. But we needed to merge this in sooner than that PR was ready because it was a blocker for others.

PR #54, opened 2018-10-28
Modified implementation of getting languages route

### LINKED ISSUE #32: Need GET /languages Route
This endpoint should get the list of languages and should implement pagination in the same manner as the patterns already implemented in the /resources endpoint (See #21)

michizhou:
Refactored the implementation for pagination of /languages route to match that of the /resources route due to changes in paginator interface. Updated code passes in query to `language_paginator.items` and removes `Language` argument from Paginator constructor. Closes #32.

### DIFF
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -117,8 +117,12 @@ def get_resources():
 
 def get_languages():
     try:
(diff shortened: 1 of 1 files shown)


aaron-suarez:
Works great, thanks for this

PR #63, opened 2018-11-05
Fix bug in PUT /resources

aaron-suarez:
This fixes a bug where if you tried to set "paid" to `false` or "notes" to `""` or `null`, it would not update the resource.


app/api/routes.py
```diff
@@ -173,10 +173,11 @@ def set_resource(id, json, db):
                 resource.name = json.get('name')
             if json.get('url'):
```
wimo7083: I'll merge this, but we should be consistent and do this for all of them.

PR #64, opened 2018-11-05
Teardown requests

### LINKED ISSUE #5: Make rollbacks a bit cleaner
Something like this 

```
@app.teardown_request
def teardown_request(exception):
    if exception:
        db.session.rollback()
    db.session.remove()
```

http://flask.pocoo.org/docs/1.0/api/#flask.Flask.teardown_request

aaron-suarez:
Fixes #5 (at least I think it does).

I need a really good code review because I don't know if I put everything in the right place. It does work, but I don't know if it's good code.

Also, sorry about the fact that there's so much more to the diff due to indentation change only. The purpose of that was to remove all of the `except` blocks because the new `teardown_request` handles the exceptions now.

@heavysaturn Would you please take a look at this too? I can't officially request a review from you, not sure why.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -27,9 +27,7 @@ def create_app(config_class=Config):
     from app.api import bp as api_bp
     app.register_blueprint(api_bp, url_prefix='/api/v1')
 
-    from app.api.errors import page_not_found
-    from app.api.errors import internal_server_error
-    app.register_error_handler(404, page_not_found)
-    app.register_error_handler(500, internal_server_error)
+    from app.errors import bp as error_bp
+    app.register_blueprint(error_bp)
 
     return app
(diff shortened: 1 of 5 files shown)


wimo7083:
I'll try and look at this tomorrow and @heavysaturn is more than welcome to check it out also. If you didn't know there's a way to [ignore whitespace](https://blog.github.com/2018-05-01-ignore-white-space-in-code-review/).

aaron-suarez:
It doesn't work with the blueprint. Besides, it won't work for the 404 route.

>the blueprint cannot handle 404 routing errors because the 404 occurs at the routing level before the blueprint can be determined.

source: http://flask.pocoo.org/docs/dev/errorhandling/

If you want to do it using blueprints, we should set up a slack call to go over it together so you can point out the specifics.

aaron-suarez:
@wimo7083 I don't know if there was an update or something, but with the current iteration of the code, 404 and 500 both work as expected. The teardown request seems to work as well. What do you think of the structure now?

Issue #25, opened 2018-10-17
Make API requests a little better (still need rate limiting)


wimo7083:
In order to prevent spam we want to setup all routes to have an api limiter:
https://flask-limiter.readthedocs.io/en/stable/

We also wish to start returning routes with more detailed information than the data in json format

If there's a style you prefer we can use it, here's one option. [google json ](https://google.github.io/styleguide/jsoncstyleguide.xml)
good response example:
```json
{
  "data": {
    "id": 1001,
    "name": "Wing"
  }
}
```

error response example:
```json
{
  "error": {
    "code": 404,
    "message": "ID not found"
  }
}
```

Some examples include, returning a status for any modifications, returning an error code if something irrecoverable happens, return a page value and in the case of lists a number of resources.

aaron-suarez:
For the responses that contain a number of resources, we should list the count of how many were returned. Theoretically, the pagination should take care of that, and whatever frontend consumes it can just do some type of `count()` or `length()` function in whatever language it's written in, but I have a feeling that having it in the response would still be useful.

heavysaturn:
A very classy way to do this (I'm so sorry!) would be to make use of class-based endpoints instead of the decorator and function-based endpoints we're using right now. 

Flask has something called MethodView that you can inherit, which allows you to write a `get()` method or a `put()` method for that class in order to handle that specific http method. So then you create a hierarchy of classes where, for example, APIView inherits MethodView and your endpoint class inherits APIView. 

What this would allow is to abstract this kind of return data formatting stuff to the APIView parent, and then the endpoints themselves could have a more abstract relationship to everything. You might raise an error simply by doing `return self.error(404, "Page not found")`, and you might use `return self.response({"name": "lemon"})`, and those methods would jsonify the data and include all the other generic information.

Does that make sense? 

It might be overcomplicating things for potential contributors, though. Thoughts?

aaron-suarez:
I like that because it reminds me of a PHP framework I used to use at my old job that would abstract that stuff away from you. You would just add your data to the response object and it would take care of the rest.

I don't think that adding endpoints would be overly complicated for beginners/potential contributors, they would just follow the patterns of the other endpoints. It might confuse them when they move onto other flask projects and that stuff they're used to using is no longer available. But honestly, every project I've ever come in on that was already started has stuff like that, so maybe that's the real lesson here.

I like the idea, but I don't want to dedicate the time to implementing it myself. If you're willing to do that part of the work, I'm ok with that.

[...]

PR #67, opened 2018-11-24
Return routes with more information about the request

aaron-suarez:
Addresses one of the points in #25.

This adds helpful information to each response, such as API version and status. Willing to have a conversation about what other information might be helpful for all requests.

### DIFF
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -7,7 +7,7 @@ from sqlalchemy.orm.exc import MultipleResultsFound, NoResultFound
 
 from app.api import bp
 from app.models import Language, Resource, Category
-from app import Config, db
+from app import Config, db, API_VERSION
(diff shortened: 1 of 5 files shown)


app/api/routes.py
```diff
@@ -39,6 +39,15 @@ def categories():
 
 
 # Helpers
+def standardize_response(data):
```
wimo7083: Possibly have this function take in a status and response_data ("data" or "error" and then the formatted data so we don't repeat our-self in error handling?

aaron-suarez: If I do that, where should that function live that it would be accessible to both `handlers.py` as well as `routes.py`?

wimo7083: maybe in utils like the paginator?

Issue #66, opened 2018-11-18
Ensure we can query by date added


aaron-suarez:
When we do the front end, we'll likely have a section for "recently added resources". We'll need a way to query for the resources to put in that section. I'm not sure if we have timestamp columns for `date_added` and `date_modified` (or similar) but we need those if we don't have them.

PR #70, opened 2018-11-25
Add create and update timestamps to Resource objects

aaron-suarez:
This addresses #66.

I am thinking we'll probably need to add a parameter to the `GET /resources` route so that the front end can actually query for resources based on the `created_at` or `last_updated` data. Let me know if you want to see that as part of this PR. 

I'm willing to have a discussion about whether we should include these timestamps in the data returned by all requests (I wasn't sure so I decided to not include it until we can have a discussion about it).

### DIFF
--- a/.env
+++ b/.env
@@ -1,3 +1,4 @@
 FLASK_APP=run.py
-SQLALCHEMY_DATABASE_URI=postgresql://user_name:change_password@127.0.0.1:5432/resources
+SQLALCHEMY_DATABASE_URI=postgresql://aaron:getmein@127.0.0.1:5432/resources
 FLASK_SKIP_DOTENV=1
+FLASK_ENV=development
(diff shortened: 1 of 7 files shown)


aaron-suarez:
Query should look like this `GET /resources?updated_after=2018-11-24`

The model itself should have something like this
```
created_at = Column(DateTime(timezone=True), server_default=func.now())
last_updated = Column(DateTime(timezone=True), onupdate=func.now())
```
That way, any changes made on the database (outside the app) will also have timestamps. See [here](https://stackoverflow.com/questions/13370317/sqlalchemy-default-datetime/33532154#33532154)

aaron-suarez:
@wimo7083 I need help writing tests for this. Can you hit me up on slack when you get a chance?

PR #73, opened 2018-12-14
Providing an update to the installation instructions

McDuckerton:
I updated the installation instructions that resolves a couple of issues:

1. Added instructions to change the pg_hba.conf file to modify the method column in order to allow users created in postgresql on linux to run createdb command.
2. Added instructions and the command to run if virtualenv is being created using python2.7 instead of the required python3.7.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,9 +81,10 @@ Note: if you try to use the same username as the user you are logged in as, you 
 SQLALCHEMY_DATABASE_URI=postgresql://aaron:password@127.0.0.1:5432/resources
 ```
 8. Navigate to the cloned repo directory `cd \path\to\resources`. (see step 6 under getting started)
-9. Install virtualenv if you do not have it. `pip install virtualenv`.
+9. Install virtualenv if you do not have it. `pip install virtualenv`.(This may need to be run with 'pip3' in place of 'pip'.)
 10. [Create a virtual environment](https://docs.python.org/3/library/venv.html) called venv `python -m virtualenv venv` (If your linux machine is creating the virtualenv with python2.X you will need to run `python -m virtualenv --python=/usr/bin/python3.7 venv`. This will force the virtualenv to be created with the minimum required version of python.)
-11. Activate virtual environment `source venv/bin/activate` (This may need to be run with 'pip3' in place of 'pip'.)
+11. Activate virtual environment `source venv/bin/activate` 
+12. Install the required dependencies 'pip install pipenv'. 
 13. Install --dev `pipenv install --dev`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -82,9 +82,8 @@ SQLALCHEMY_DATABASE_URI=postgresql://aaron:password@127.0.0.1:5432/resources
 ```
 8. Navigate to the cloned repo directory `cd \path\to\resources`. (see step 6 under getting started)
 9. Install virtualenv if you do not have it. `pip install virtualenv`.
-10. [Create a virtual environment](https://docs.python.org/3/library/venv.html) called venv `python -m virtualenv venv`.
-11. Activate virtual environment `source venv/bin/activate`
-12. Install the required dependencies `pip install pipenv`.
+10. [Create a virtual environment](https://docs.python.org/3/library/venv.html) called venv `python -m virtualenv venv` (If your linux machine is creating the virtualenv with python2.X you will need to run `python -m virtualenv --python=/usr/bin/python3.7 venv`. This will force the virtualenv to be created with the minimum required version of python.)
+11. Activate virtual environment `source venv/bin/activate` (This may need to be run with 'pip3' in place of 'pip'.)
```
aaron-suarez: I think this note about pip3 was supposed to go on number 9, when the command is `pip install virtualenv` (which should, of course, be `pip3 install virtualenv`)

README.md
```diff
@@ -110,7 +109,7 @@ SQLALCHEMY_DATABASE_URI=postgresql://aaron:password@127.0.0.1:5432/resources
...
-5. Create a database with `sudo -u postgres createdb resources -U aaron` ("resources" is the name we are giving to the database. You can call it whatever you'd like. Make sure to replace "aaron", with whatever username you gave to your postgres user.)
+5. Create a database with `sudo -u postgres createdb resources -U aaron` ("resources" is the name we are giving to the database. You can call it whatever you'd like. Make sure to replace "aaron", with whatever username you gave to your postgres user.). **Note: In order for this to work on linux/unix you must change the "method" column for all types to "trust" in the pg__hba.conf file most likely located here: ```/etc/postgresql/9.x/main/pg_hba.conf```.
```
aaron-suarez: Nit: you only need one set of backticks here. Also, should we maybe make this comment bold by ending with some asterisks? What do you think of this
```
`/etc/postgresql/9.x/main/pg_hba.conf`**.
```

README.md
McDuckerton: Yes! I just missed that one.

README.md
McDuckerton: That looks good. I think the firectory should be /10/ in place of /9.x/. This was a copy/paste I forgot to edit I think

PR #77, opened 2018-12-24
Add coveralls to track code coverage

aaron-suarez:
Now that we have some working tests via #70, we should also add code coverage via coveralls. Ideally, we'll be able to increase our code coverage to something very close to 100%, but we'll see once we get there. For now, let's just go ahead and get tracking on it.


aaron-suarez:
Close in favor of #80

PR #84, opened 2018-12-29
Add rate limiting

### LINKED ISSUE #25: Make API requests a little better (still need rate limiting)

aaron-suarez:
Fixes #25 

Adds rate limiting and a test for it (even though I know you said that we probably don't need to test this, I still feel like it's warranted as a regression prevention mechanism).


app/errors/handlers.py
```diff
@@ -14,7 +14,19 @@ def page_not_found(e):
     ]
 
     # Set the 404 status explicitly
-    return standardize_response(None, errors, "not found"), 404
+    return standardize_response(None, errors, "not found", 404)
+
+
+@bp.app_errorhandler(429)
```
wimo7083: I assume that flask-limiter raises a 429 which is how this handler gets called.

tests/unit/test_routes.py
```diff
@@ -50,6 +50,18 @@ def test_update_resource(app, session, db):
     assert(response.json['data'].get('name') == "New name")
 
 
+def test_rate_limit(app, session, db):
+    client = app.test_client()
```
wimo7083: TODO: make this faster, test are currently at 20 seconds. Not sure how long this takes.

app/errors/handlers.py
aaron-suarez: Yes, flask-limiter raises a 429 which returns HTML to the client. This method returns JSON instead

tests/unit/test_routes.py
aaron-suarez: It takes less than 20 seconds on my machine... I think the longest part is the `test_getters` method, which makes sense because it has to import all the resources from resources.yml first. See the tests run with the duration flag:

```
$ pipenv run pytest --durations=0
Loading .env environment variables…
====================================================================== test session starts ======================================================================
platform darwin -- Python 3.7.0, pytest-4.0.2, py-1.7.0, pluggy-0.8.0
rootdir: /Users/aaronsuarez/Github/resources_api, inifile: pytest.ini
plugins: cov-2.6.0
collected 4 items

tests/unit/test_routes.py ....                                                                                                                            [100%]

==================================================================== slowest test durations =====================================================================
2.28s call     tests/unit/test_routes.py::test_getters
0.06s call     tests/unit/test_routes.py::test_rate_limit
0.03s setup    tests/unit/test_routes.py::test_getters
0.01s call     tests/unit/test_routes.py::test_update_resource
0.01s call     tests/unit/test_routes.py::test_create_resource

(0.00 durations hidden.  Use -vv to show these durations.)
=================================================================== 4 passed in 2.58 seconds ====================================================================
```

PR #86, opened 2019-01-02
Add .yml file to ignore specified folders and files

KPGunner:
Created codeclimate.yml file to ignore "migrations" folder from Code Climate checks.


aaron-suarez:
Oh no! https://codeclimate.com/github/OperationCode/resources_api/builds/34

@KPGunner Can you open another PR to fix the error?

PR #89, opened 2019-01-08
Update .codeclimate.yml

KPGunner:


KPGunner:
Awesome. Glad it worked. I should have spent some more time looking at the documentation and I would have caught it on the first run through.

Issue #91, opened 2019-01-10
/static/{filename} route should be removed


aaron-suarez:
By default, flask has a route called `static` that lets you get a static file. This is a JSON API, there is no sense in allowing users to `GET` static files. Therefore, the default `/static` route should be removed.

aaron-suarez:
Assigned myself as a placeholder for @KPGunner

aaron-suarez:
Closed by #102

PR #94, opened 2019-01-10
Removed static route

KPGunner:


aaron-suarez:
@KPGunner This doesn't actually solve the problem specified by #91. As a matter of fact, we actually want to keep the favicon, but that's a different route.

If you run `make routes`, this is the output:
```
docker-compose run resources-api flask routes
Starting resources-psql ... done
Endpoint             Methods  Rule
-------------------  -------  --------------------------
api.apikey           POST     /api/v1/apikey
api.create_resource  POST     /api/v1/resources
api.get_categories   GET      /api/v1/categories
api.get_languages    GET      /api/v1/languages
api.get_resource     GET      /api/v1/resources/<int:id>
api.get_resources    GET      /api/v1/resources
api.update_resource  PUT      /api/v1/resources/<int:id>
static               GET      /static/<path:filename>
views.favicon        GET      /favicon.ico
```
The goal is to remove this route, not the favicon. If necessary, you can just change the name of the directory to `icons` and change app/views/routes.py to reference `icons` instead of `static`. But that still doesn't remove the route.

KPGunner:
Okay. I think I figured out a solution. I'll try to get to this today.

KPGunner:
I think I got it this time. Make routes returns favicon without static route.

PR #95, opened 2019-01-19
Add authentication to certain routes

### LINKED ISSUE #52: Need authentication and/or data rollback mechanism
Once we merge #38, if we were to put this in production, anyone could change any resource to point to anything. Though we do indeed want to provide functionality to be able to add notes, update URLs, etc, we do not want trolls and hackers to be able to corrupt all of our data. We need a way to review changes as they're coming in so that if bogus or incorrect data is being injected to our database, we're aware of it and can fix it. And preferably, only authenticated users should be able to update, and we should also be able to lock out users who prove that they'd rather troll than provide value to our resource collection.

I know that we were planning on having the front end authenticate users using the same methods as the Operation Code backend. I think there were also talks of pybot/pyback being able to reuse authentication credentials for various services.

We need to hash out exactly what this is going to look like and implement this mechanism before we put this in production.

aaron-suarez:
Fixes #52

I couldn't figure out how to fix the merge conflicts on #92 so I went ahead and branched from master and created this new PR.


app/api/routes.py
```diff
@@ -42,6 +49,26 @@ def categories():
     return get_categories()
 
 
+@bp.route('/apikey', methods=['POST'], endpoint='apikey')
```
wimo7083: Is this a good name for this route? What are the conventions out there?

tests/unit/test_persistent_routes.py
```diff
@@ -1,124 +0,0 @@
-import pytest
-from app.models import Resource, Language, Category
-from tests.conftest import module_client, module_db
-from configs import PaginatorConfig
-
-
-# TODO: We need negative unit tests (what happens when bad data is sent)
-def test_get_resources(module_client, module_db):
```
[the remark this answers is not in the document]
aaron-suarez: Just to be clear, the reason I deleted `test_persistent_routes.py` is because with authentication required, we need to store the auth token in the database upon successful authentication, and that made every single "non-persistent" route persistent. Therefore, if every route needs a persistent database, then they're all the same and we may as well just call it `test_routes.py`. Technically, we don't need to have all the resources imported for every route, but we do need to have the auth token stored in the DB for create/update.

If there's a better way, let me know, but honestly, at this point I'm a fan of having all the routes be tested using the same test db

app/api/routes.py
```diff
@@ -176,27 +207,32 @@ def get_attributes(json):
 def set_resource(id, json, db):
     resource = None
     resource = Resource.query.get(id)
+    logger.info(f"Updating resource. Old data: {resource.serialize}")
```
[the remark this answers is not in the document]
aaron-suarez: To answer your question literally, this is what happens:
```
AttributeError: 'NoneType' object has no attribute 'serialize'
```
But what's supposed to happen is this:
https://github.com/OperationCode/resources_api/blob/44b42f29a7073befb1ae44eb5d229bac48bb6511/app/api/routes.py#L267
I'll put that inside the `try/catch`

app/api/routes.py
aaron-suarez: I think the convention is to get an API key from a website, rather than from the API itself. For example, from https://docs.rapidapi.com/docs/keys

screen shot 2019-02-11 at 8 51 17 pm

After much searching, I found that Sendgrid calls theirs `/api_key`. See https://sendgrid.api-docs.io/v3.0/api-keys/create-api-keys

Would you like me to add an underscore?

app/api/routes.py
```diff
@@ -4,32 +4,39 @@
...
+@bp.route('/resources/<int:id>', methods=['GET'], endpoint='get_resource')
 def resource(id):
-    if request.method == 'GET':
-        return get_resource(id)
-    elif request.method == 'PUT':
-        return set_resource(id, request.get_json(), db)
+    return get_resource(id)
+
+
+@bp.route('/resources/<int:id>', methods=['PUT'], endpoint='update_resource')
```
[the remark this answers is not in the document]
aaron-suarez: I'm not sure I understand what you mean. If you try to update a resource that doesn't exist, you should either get a 400 or a 404. Unless we want to let `PUT` and `POST` be interchangeable (i.e., create if it doesn't exist, update if it does). But I'm not a big fan of the latter, I would prefer an error if they try to update something that doesn't exist

app/api/routes.py
```diff
@@ -42,6 +49,26 @@ def categories():
...
+@bp.route('/apikey', methods=['POST'], endpoint='apikey')
+def apikey():
+    json = request.get_json()
+    user = auth.check_user_with_oc(json)
+
+    if not user:
+        errors = [{"code": "not-authorized"}]
+        return standardize_response(None, errors, "not authorized", 401)
+
+    try:
```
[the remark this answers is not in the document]
aaron-suarez: Wait, what? If the token is found, we need to use it to find the apikey. Are we caching things like this? If so, can you tell me the syntax to take advantage of it? If not, I'll need you to expand on how exactly to implement caching

app/api/routes.py
```diff
@@ -42,6 +49,26 @@ def categories():
...
+        apikey = Key.query.filter_by(email=json.get('email')).first()
+        if not apikey:
+            return create_new_apikey(json.get('email'))
```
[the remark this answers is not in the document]
aaron-suarez: Oh were you just stating what's happening? I think the code is pretty readable and speaks for itself, but if you think we need comments (or to refactor for readability) let me know

app/models.py
```diff
@@ -36,6 +36,15 @@ class Resource(db.Model):
     @property
     def serialize(self):
         """Return object data in easily serializeable format"""
+        if self.created_at:
```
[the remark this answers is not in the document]
aaron-suarez: That looks like a fairly big lift. If you wouldn't mind, I'd like to open that as a separate issue, and just go ahead and get this merged without that change

app/models.py
```diff
@@ -136,3 +147,28 @@ def __hash__(self):
...
+class Key(db.Model):
+    id = db.Column(db.Integer, primary_key=True)
+    apikey = db.Column(db.String, unique=True, nullable=False)
+    email = db.Column(db.String, unique=True, nullable=False)
+
+    @property
+    def serialize(self):
+        """Return object data in easily serializeable format"""
+        return {
+            'apikey': self.apikey,
```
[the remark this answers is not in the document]
aaron-suarez: It works fine when printed or logged... Why do we need to call `str()` on it?

app/api/routes.py
```diff
@@ -244,7 +285,28 @@ def create_resource(json, db):
...
-    db.session.add(new_resource)
-    db.session.commit()
+    try:
+        db.session.add(new_resource)
+        db.session.commit()
+
+        return standardize_response(new_resource.serialize, None, "ok")
+    except Exception as e:
+        logger.error(e)
+        return standardize_response(None, [{"code": "bad-request"}], "bad request", 400)
```
aaron-suarez: TODO: I would like to change these to `redirect('/400')` and put in a generic `400` handler. Also, might want to look into returning validation errors as `422` rather than `400`, and a generic exception here might actually need to be a 500...

This PR is growing and growing, and I don't want to keep growing it by adding all that in, so I'd rather roll with it as is and open those as separate issues that are beginner friendly

docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
     env_file:
     - .env
     volumes:
-    - .:/app
+    - .:/src
```
aaron-suarez: The docker container can't currently write to the filesystem. This fixes that. This should really be it's own PR....

tests/unit/test_routes.py
```diff
@@ -9,54 +9,231 @@
...
+    return client.post('/api/v1/resources',
+        json={
+            "name": "Some Name",
+            "url": "http://example.org/",
+            "category": "New Category",
+            "languages": ["Python", "New Language"],
+            "paid": False,
+            "notes": "Some notes"
+        },
+        headers={'x-apikey': apikey}
```
aaron-suarez: These ones need to be in this format because if you do this
```python
headers = dict(x-apikey=apikey)
```
you get this:
```SyntaxError: keyword can't be an expression```

File: .env [hunk 12]
```diff
@@ -13,3 +13,4 @@ PASSWORD=change_password
 DB=resources-psql
 FLASK_SKIP_DOTENV=1
 FLASK_ENV=development
+FLASK_DEBUG=1
```
aaron-suarez: This, in conjunction with the change from `app` to `src` in docker-compose.yml allows live-reloading while developing

tests/unit/test_routes.py
```diff
@@ -9,54 +9,231 @@
...
+
+    # Test bigger than max page size
+    too_long = PaginatorConfig.max_page_size + 1
+    response = client.get(f"api/v1/resources?page_size={too_long}")
+    assert (len(response.json['data']) == PaginatorConfig.max_page_size)
+
+    # Test farther than last page
+    too_far = 99999999
+    response = client.get(f"api/v1/resources?page_size=100&page={too_far}")
+    assert (len(response.json['data']) == 0)
```
aaron-suarez: Should this be an error (404, perhaps?) instead of returning an empty list? We may have already talked about this, but I feel like it's worth reconsidering, especially for consistency across the API. If you request an ID that's too high, you get a 404. If you request a page that's too high, maybe you also should get a 404?

tests/unit/test_routes.py
```diff
@@ -9,54 +9,231 @@
...
+
+def test_languages(module_client, module_db):
+    client = module_client
 
-    response = create_resource(client)
+    response = client.get('api/v1/languages')
+    for language in response.json['data']:
+        assert (isinstance(language.get('id'), int))
+        assert (isinstance(language.get('name'), str))
+        assert (len(language.get('name')) > 0)
```
aaron-suarez: In some tests we check whether strings are empty with the following syntax:
```python
assert (resource.get('name') != "")
```
We should really pick one (check if length is 0 or check if value == "") and use that same syntax throughout the tests

app/api/routes.py
```diff
@@ -52,6 +59,27 @@ def categories():
     return get_categories()
 
 
+@bp.route('/apikey', methods=['POST'], endpoint='apikey')
+def apikey():
+    json = request.get_json()
+    token = check_user_with_oc(json)
```
aaron-suarez: pass in user and password rather than the whole json

aaron-suarez: call "token" `is_oc_member`

aaron-suarez: `is_oc_user: bool = check_user_with_oc(....)`

app/api/routes.py
```diff
@@ -52,6 +59,27 @@ def categories():
...
+
+    if not token:
+        errors = [{"code": "not-authorized"}]
+        return standardize_response(None, errors, "not authorized", 401)
+
+    try:
+        apikey = Key.query.filter_by(email=json.get('email')).first()
```
aaron-suarez: Leave a comment about what we're doing in this flow

app/api/routes.py
```diff
@@ -244,7 +285,28 @@ def create_resource(json, db):
...
+
+        return standardize_response(new_resource.serialize, None, "ok")
+    except Exception as e:
+        logger.error(e)
+        return standardize_response(None, [{"code": "bad-request"}], "bad request", 400)
+
 
-    return standardize_response(new_resource.serialize, None, "ok")
+def create_new_apikey(email):
+    new_key = Key(
```
aaron-suarez: TODO: create the key in a while loop to ensure that accidental duplicates are caught and handled (in the extremely unlikely chance that there's a collision).

app/models.py
```diff
@@ -136,3 +147,28 @@ def __hash__(self):
 
     def __repr__(self):
         return f"<Language {self.name}>"
+
+
+class Key(db.Model):
+    id = db.Column(db.Integer, primary_key=True)
+    apikey = db.Column(db.String, unique=True, nullable=False)
+    email = db.Column(db.String, unique=True, nullable=False)
```
aaron-suarez: Add created/updated in the Key class

aaron-suarez: Add index on apikey

tests/conftest.py
```diff
@@ -85,3 +85,17 @@ def module_db():
     yield _db  # this is where the testing happens!
 
     _db.drop_all()
+
+@pytest.fixture(scope='function')
+def patched_request(mocker):
+    class MyRequest(object):
```
aaron-suarez: Rename to explain what these specific requests are for

aaron-suarez: Put a comment in there too to explain what's going on

tests/conftest.py
```diff
@@ -85,3 +85,17 @@ def module_db():
     yield _db  # this is where the testing happens!
 
     _db.drop_all()
+
+@pytest.fixture(scope='function')
+def patched_request(mocker):
```
aaron-suarez: Rename to make it clear that it should only apply when you need a token from the OC backend API

tests/unit/test_routes.py
```diff
@@ -9,54 +9,232 @@
...
+    client = module_client
+
+    apikey = get_api_key(client)
+    response = create_resource(client, apikey)
     assert (response.status_code == 200)
     assert (isinstance(response.json['data'].get('id'), int))
     assert (response.json['data'].get('name') == "Some Name")
 
 
-def test_update_resource(session_app, function_session, session_db):
```
aaron-suarez: These three functions should be marked with a comment that they aren't currently in use but we'd like to go back to using them when we can avoid _always_ using the persistent DB

app/api/auth.py
```diff
@@ -0,0 +1,42 @@
...
+            errors = [{"code": "not-authorized"}]
+            return standardize_response(None, errors, "not authorized", 401)
+
+        log_request(request, key)
+
+        return func(*args, **kwargs)
+    return wrapper
+
+
+def check_user_with_oc(json):
```
aaron-suarez: change to `is_user_oc_member` to indicate it should return a boolean

Issue #97, opened 2019-01-26
`standardize_response` is awkward


aaron-suarez:
I like the idea of having a function to standardize all of the responses so we can respond in certain ways without having to think too hard about how to structure the response while maintaining a consistent API experience.

However, the interface for the current implementation is pretty awful. At the very least, it should use keyword arguments so that it's obvious which parameter is which when you're reading through the code. But, perhaps with a little thought, a solution that's a bit less clunky and more readable can be designed.

I'm open to suggestions about how this might look.

mildocjr:
It seems to be a typical format, however, something I do is pass a dictionary in the format `{'data': data, 'errors': [errors]}` to a `payload` argument and let the browser handle the OK status.  Then you could pull out each of these in the `standardize_response` function, and determine an appropriate error code which could be overridden if needed as a named argument.

General format would be:
```python
def standarize_response(payload, status_code=200):
    # pull out possible values from payload or set to some default value 
    data = payload["data"] if 'data' in payload else None     # maybe use {} instead
    errors = payload["errors"] if 'errors' in payload else []

    if len(data):
        resp["data"] = data
        status = "ok"         # might just make this an dict keyed by status_code

    elif len(errors):
        resp["errors"] = errors
        if status_code == 400:
            status = "bad request"
        elif status_code in [401, 403]:
            status = "unauthorized"
        elif status_code == 404:
            status = "not found"          # dict keyed by status_code would be better

    else:
        resp["errors"] = [{"code": "something-went-wrong"}] # will this handle 500 errors?

    resp = {
        "status": status,
        "apiVersion": API_VERSION
        "data": data        # always include data even if its just the original POST form
        "errors" errors     # always include errors even if there are none
    }

    return jsonify(resp), status_code
```

Then the call would just be simple
```python
# success
standardize_response(data)

# failure
standardize_response(data, 404)
```

Not sure if this is what you had in mind but it's one option, another is to pass data and errors in as a named tuple.

aaron-suarez:
You know, I forgot that keyword arguments are a thing in python, I may think about doing it that way... Thanks for the suggestion, I'm sure it will ultimately influence the final implementation.

If you wanted to open up a PR to fix it, that's fine too! Contributors wanted 😄

aaron-suarez:
Closed in #110

PR #110, opened 2019-03-11
proposed solution to #97

lord-cant-even:


app/api/routes.py
```diff
@@ -106,9 +104,9 @@ def get_resource(id):
 
     finally:
         if resource:
-            return standardize_response(resource.serialize, None, "ok")
+            return standardize_response(payload=dict(data=(resource.serialize)))
         else:
-            return standardize_response({}, None, "ok")
+            return standardize_response(payload=dict(data={}))
```
aaron-suarez: I realized a while ago that this is actually incorrect, we should return a 404 if the resource is not found in the DB. Do you think you can go ahead and update that while you're in there?

app/utils.py
```diff
@@ -18,18 +18,60 @@ def items(self, query):
...
+            resp["errors"] = errors
+        else:
+            resp["errors"] = [{
+                "code": '-'.join(err_map.get(status_code).split(' ')).lower()
+            }]
+
+    elif not data:
         resp["errors"] = [{"code": "something-went-wrong"}]
-    return jsonify(resp), status_code
+        resp["status_code"] = 500
```
aaron-suarez: In the situation where there is no data, we need to also set the `status_code` variable to `500`. Right now, when I get a 500 error with code `something-went-wrong` it's returning a 200 status with the 500 JSON.

To try it out, do `GET /api/v1/resources/0` (there is no resource with ID 0, so it errors out). You'll see the 200 response with the 500 JSON

app/api/routes.py
```diff
@@ -244,13 +242,13 @@ def update_votes(id, vote_direction):
...
         print(e)
-        return standardize_response(None, None, None, 500)
+        return standardize_response(payload=None, status_code=500)
 
     initial_count = getattr(resource, vote_direction)
     setattr(resource, vote_direction, initial_count+1)
     db.session.commit()
 
-    return standardize_response(resource.serialize, None, "ok")
+    return standardize_response(payload=dict(data=resource.serialize), status_code=200)
```
aaron-suarez: Since the default for `status_code` is 200, I don't think we need that to be specified here

app/api/routes.py
```diff
@@ -205,7 +203,7 @@ def get_categories():
         print(e)
         category_list = []
     finally:
-        return standardize_response(category_list, None, "ok")
+        return standardize_response(payload=dict(data=category_list), status_code=200)
```
aaron-suarez: Since the default for `status_code` is 200, I don't think we need that to be specified here

app/api/routes.py
```diff
@@ -186,9 +184,9 @@ def get_languages():
         ]
     except Exception as e:
         logger.error(e)
-        return standardize_response(None, [{"code": "bad-request"}], "bad request", 400)
+        return standardize_response(payload=None, status_code=400)
 
-    return standardize_response(language_list, None, "ok")
+    return standardize_response(payload=dict(data=language_list), status_code=200)
```
aaron-suarez: Since the default for `status_code` is 200, I don't think we need that to be specified here

app/api/routes.py
```diff
@@ -171,9 +169,9 @@ def get_resources():
         ]
     except Exception as e:
         logger.error(e)
-        return standardize_response(None, [{"code": "bad-request"}], "bad request", 400)
+        return standardize_response(payload=None, status_code=400)
 
-    return standardize_response(resource_list, None, "ok")
+    return standardize_response(payload=dict(data=resource_list), status_code=200)
```
aaron-suarez: Since the default for `status_code` is 200, I don't think we need that to be specified here

app/api/routes.py
```diff
@@ -171,9 +169,9 @@ def get_resources():
         ]
     except Exception as e:
         logger.error(e)
-        return standardize_response(None, [{"code": "bad-request"}], "bad request", 400)
+        return standardize_response(payload=None, status_code=400)
```
aaron-suarez: I'm on the fence about whether this should actually be a `400` or `500` status code... If an exception occurs here, that's unexpected, right? Which means that it's an unexpected server error, which would be a 500 response....

And this is in `get_resources` right? So if they get this far, then it's not a 404 by any means... It should probably be a `500`. Open to discussion on this point

app/api/routes.py
```diff
@@ -279,10 +277,13 @@ def set_resource(id, json, db):
...
 
-        return standardize_response(resource.serialize, None, "ok")
+        return standardize_response(
+            payload=dict(data=resource.serialize),
+            status_code=200
+            )
     except Exception as e:
         logger.error(e)
-        return standardize_response(None, [{"code": "bad-request"}], "bad request", 400)
+        return standardize_response(payload=None, status_code=400)
```
aaron-suarez: Same here, I think maybe this should be a `500` instead of a `400`

app/api/routes.py
```diff
@@ -299,10 +300,13 @@ def create_resource(json, db):
...
 
-        return standardize_response(new_resource.serialize, None, "ok")
+        return standardize_response(
+            payload=dict(data=new_resource.serialize),
+            status_code=200
+            )
     except Exception as e:
         logger.error(e)
-        return standardize_response(None, [{"code": "bad-request"}], "bad request", 400)
+        return standardize_response(payload=None, status_code=400)
```
aaron-suarez: Same here, `500`

app/utils.py
```diff
@@ -18,18 +18,60 @@ def items(self, query):
         return query.paginate(self.page, self.page_size, False).items
 
 
-def standardize_response(data, errors, status, status_code=200):
-    resp = {
-        "status": status,
-        "apiVersion": API_VERSION
+def standardize_response(payload, status_code=200):
```
aaron-suarez: What do you think about making the default for payload `None`, so it would look like 
```python
def standardize_response(payload=None, status_code=200)
```
Then, when you don't have a payload (e.g., when doing a 401 unauthorized), you can just say `standardize_response(status_code=401)`?

app/api/routes.py
```diff
@@ -186,9 +184,9 @@ def get_languages():
         ]
     except Exception as e:
         logger.error(e)
-        return standardize_response(None, [{"code": "bad-request"}], "bad request", 400)
+        return standardize_response(payload=None, status_code=400)
```
aaron-suarez: Same here, `500` error

app/api/routes.py
```diff
@@ -279,10 +277,13 @@ def set_resource(id, json, db):
 
         db.session.commit()
 
-        return standardize_response(resource.serialize, None, "ok")
+        return standardize_response(
+            payload=dict(data=resource.serialize),
+            status_code=200
```
aaron-suarez: 200 unnecessary here

app/api/routes.py
```diff
@@ -299,10 +300,13 @@ def create_resource(json, db):
         db.session.add(new_resource)
         db.session.commit()
 
-        return standardize_response(new_resource.serialize, None, "ok")
+        return standardize_response(
+            payload=dict(data=new_resource.serialize),
+            status_code=200
```
aaron-suarez: Same here, unnecessary 200

app/api/routes.py
```diff
@@ -319,8 +323,10 @@ def create_new_apikey(email):
         db.session.add(new_key)
         db.session.commit()
 
-        return standardize_response(new_key.serialize, None, "ok")
+        return standardize_response(
+            payload=dict(data=new_key.serialize),
+            status_code=200
```
aaron-suarez: here too

lord-cant-even:
Updates pushed! :)

Issue #79, opened 2018-12-24
Test coverage


aaron-suarez:
We are at about 66% test coverage as of today. It would be nice to be able to be somewhere >90% code coverage, or even 100% if possible (though it may not be). One of the specific points that is currently missing a test is surrounding filtering `GET /resources` by `updated_after` query param.

aaron-suarez:
![Screen Shot 2019-03-25 at 10 53 59 AM](https://user-images.githubusercontent.com/2008701/54934344-59f53000-4eec-11e9-98e5-20601d10c244.png)

The above screenshot is current coverage after #111 was merged. It would be nice to get some more coverage around the CLI before we close this (if done comprehensively, I believe that will push us over 90% because there are 35 untested statements in that file). If whoever picks this up would like to increase coverage in the other files, please feel free!

I did everything I can do at this point in my PR, so I'll leave this open for someone else. I might circle back around to it if enough time passes...

aaron-suarez:
We're at 98% as of #142 so I'm going to go ahead and close this. Thanks!

PR #111, opened 2019-03-11
Add test coverage

aaron-suarez:
There is no test coverage asserting the JSON returned from a 500 error, so I wanted to go ahead and add that. Current implementation fails the tests rather than allowing the API to return a 500. Hopefully we can figure out how to "raise an error" without letting it bubble all the way up to the test suite.

I went ahead and tried to increase test coverage in all the ways I know how. It's now up to 88%. Suggestions about how exactly to increase test coverage in the remaining 12% would be appreciated, but I'm not sure how to test the CLI or the error handlers (as a matter of fact, I'm not entirely convinced the handlers even work, so any thoughts surrounding that are appreciated).

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,18 +1,9 @@
-from sys import version_info
-
 from configs import Config
 from flask import Flask
 from flask_migrate import Migrate
 from flask_sqlalchemy import SQLAlchemy
 from flask_limiter import Limiter
 from flask_limiter.util import get_remote_address
-
-try:
-    assert version_info >= (3, 7, 0)
-except AssertionError:
-    print('Warning Current Python version not supported')
-    print('Current Python version: {version}'.format(version=version_info))
-    exit(1)
 
 
 db = SQLAlchemy()
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -85,7 +85,7 @@ def apikey():
         logger.info(apikey.serialize)
         return standardize_response(payload=dict(data=apikey.serialize))
     except Exception as e:
-        logger.error(e)
+        logger.exception(e)
         return standardize_response(status_code=500)
 
 
@@ -97,11 +97,11 @@ def get_resource(id):
 
     except MultipleResultsFound as e:
         print_tb(e.__traceback__)
-        logger.error(e)
+        logger.exception(e)
 
     except NoResultFound as e:
         print_tb(e.__traceback__)
-        logger.error(e)
+        logger.exception(e)
         return redirect('/404')
 
     if resource:
@@ -152,7 +152,7 @@ def get_resources():
                 raise Exception("updated_after greater than today's date")
             uaDate = uaDate.strftime("%Y-%m-%d")
         except Exception as e:
-            logger.error(e)
+            logger.exception(e)
             message = 'The value for "updated_after" is invalid'
             res = dict(errors=[{'code': 'unprocessable-entity', 'message': message}])
             return standardize_response(payload=res, status_code=422)
@@ -169,7 +169,7 @@ def get_resources():
             resource.serialize for resource in resource_paginator.items(q)
         ]
     except Exception as e:
-        logger.error(e)
+        logger.exception(e)
         return standardize_response(status_code=500)
 
     return standardize_response(payload=dict(data=resource_list))
@@ -184,7 +184,7 @@ def get_languages():
             language.serialize for language in language_paginator.items(query)
         ]
     except Exception as e:
-        logger.error(e)
+        logger.exception(e)
         return standardize_response(status_code=500)
 
     return standardize_response(payload=dict(data=language_list))
@@ -200,7 +200,7 @@ def get_categories():
         ]
 
     except Exception as e:
-        logger.error(e)
+        logger.exception(e)
         return standardize_response(status_code=500)
 
     return standardize_response(payload=dict(data=category_list))
@@ -232,16 +232,16 @@ def update_votes(id, vote_direction):
 
     except MultipleResultsFound as e:
         print_tb(e.__traceback__)
-        logger.error(e)
+        logger.exception(e)
 
     except NoResultFound as e:
         print_tb(e.__traceback__)
-        logger.error(e)
+        logger.exception(e)
         return redirect('/404')
 
     except Exception as e:
         print_tb(e.__traceback__)
-        logger.error(e)
+        logger.exception(e)
         return standardize_response(status_code=500)
 
     initial_count = getattr(resource, vote_direction)
@@ -281,7 +281,7 @@ def set_resource(id, json, db):
             payload=dict(data=resource.serialize)
             )
     except Exception as e:
-        logger.error(e)
+        logger.exception(e)
         return standardize_response(status_code=500)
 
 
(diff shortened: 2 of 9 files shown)


app/__init__.py
```diff
@@ -1,19 +1,10 @@
...
-
 from configs import Config
 from flask import Flask
 from flask_migrate import Migrate
 from flask_sqlalchemy import SQLAlchemy
 from flask_limiter import Limiter
 from flask_limiter.util import get_remote_address
 
-try:
-    assert version_info >= (3, 7, 0)
```
aaron-suarez: We use docker now, which should specify the version of Python that we're using. This is now unnecessary

aaron-suarez:
Fixes #79

tests/conftest.py
```diff
@@ -114,3 +114,52 @@ def json(self):
...
+
+@pytest.fixture(scope='function')
+def fake_key_query_error(mocker):
+    """
+    Intentionally raises an error to test error response is correct
+    """
+    def key_query():
+        raise Exception()
+
+    mocker.patch('app.models.Key.query', return_value=key_query, side_effect=key_query)
```
wimo7083: Do you know if the solution was the rvalue or the side_effect? I saw things like this about using a side effect to get the mocker to see the exception.

tests/unit/test_routes.py
```diff
@@ -32,6 +34,19 @@ def test_get_resources(module_client, module_db):
...
+    uaString = ua.strftime('%m-%d-%Y')
+    response = client.get(f"/api/v1/resources?updated_after={uaString}")
+    assert (response.status_code == 200)
+
+    # If updated_after date is after today, then should return a 422
+    ua = datetime.now() + timedelta(days=1)
+    uaString = ua.strftime('%m-%d-%Y')
+    response = client.get(f"/api/v1/resources?updated_after={uaString}")
+    assert (response.status_code == 422)
+    assert (response.json.get('errors')[0].get('code') == "unprocessable-entity")
```
wimo7083: split this out into separate methods.

I'd suggest splitting this out into named methods
`test_get_resources_post_date_failure` 

Each test should ideally minimize the number of assertions so when one fails we can tell why.

tests/unit/test_routes.py
```diff
@@ -32,6 +34,19 @@ def test_get_resources(module_client, module_db):
...
+    assert ("updated_after" in response.json.get('errors')[0].get('message'))
```
wimo7083: This is tieing the implementation to the behavior too strongly. We want to _eventually_ test that on an error, we call the standard error handler. Not that the value returned is the correct format. For that we'd want to use approval tests.

tests/unit/test_routes.py
```diff
@@ -172,29 +208,101 @@ def test_get_api_key(module_client, module_db, fake_auth_from_oc):
...
+        password="supersecurepassword"
+    ))
+    assert (response.status_code == 500)
+
+def test_internal_server_error_handler(module_client, module_db, fake_items_error):
+    client = module_client
+
+    response = client.get('api/v1/resources')
+    assert (response.status_code == 500)
+    assert (response.json['errors'][0].get("code") == "server-error")
```
wimo7083: Instead of testing the json returned in many places we should be testing that the error handler is called and then separately testing the functionality of the error handler.

tests/unit/test_models.py
```diff
@@ -0,0 +1,91 @@
...
+
+def test_key():
+    key = Key(email="test@example.org", apikey="1234abcd")
+    assert (key.__hash__() == hash('1234abcd'))
+    assert (key.__repr__() == f"<Key email=test@example.org apikey=1234abcd>")
+    assert (key != 1)
+    assert (key == key)
+    time = datetime.now()
+    key.last_updated = time
+    assert (key.serialize == {'apikey': '1234abcd',
```
wimo7083: assert is an expression, should be 

`assert key.hash__() == hash('1234abcd')`
Can be done later.

tests/conftest.py
```diff
@@ -114,3 +114,52 @@ def json(self):
             return {'token': 'superlegittoken'}
 
     mocker.patch("requests.post", return_value=FakeExternalResponse())
+
+@pytest.fixture(scope='function')
+def fake_invalid_auth_from_oc(mocker):
+    """
+    Changes the return value of requests.post to be a custom response
```
wimo7083: I like this description.

tests/conftest.py
```diff
@@ -114,3 +114,52 @@ def json(self):
...
+        def json(self):
+            # Mock an error returned from the OC backend
+            return {'error': 'Invalid Email or password.'}
+
+    mocker.patch("requests.post", return_value=FakeExternalResponse())
+
+@pytest.fixture(scope='function')
+def fake_items_error(mocker):
+    """
+    Intentionally raises an error to test error response is correct
```
wimo7083: These descriptions should be more like the above one. Why do we mock this method, what are we mocking. 

If the mock is specific to a single usage, I'm unsure if intended usage should be included.

app/__init__.py
wimo7083: Make sense.

app/api/routes.py
```diff
@@ -199,11 +200,10 @@ def get_categories():
         ]
 
     except Exception as e:
-        print_tb(e.__traceback__)
-        print(e)
-        category_list = []
-    finally:
-        return standardize_response(payload=dict(data=category_list))
+        logger.error(e)
```
wimo7083: I prefer `logger.exception(e)`

tests/conftest.py
aaron-suarez: I don't know. I know that return_value by itself didn't seem to work, but I changed so many things tinkering with it that I'm not sure what actually pushed it over the edge. I can try taking away return_value and seeing if it still works

tests/unit/test_routes.py
aaron-suarez: Ok, so should I just get rid of that assertion altogether? If not, how would you change it?

app/api/routes.py
aaron-suarez: I didn't know `logger.exception()` was a method. I'm used to the syslog levels of logging (Alert, Critical, Error, Warning, Notice, Informational, Debug) and the related methods associated (i.e., `error()`, `warn()`, `debug()`). I'll have to read up on logger.exception to see what the differences are.

tests/conftest.py
aaron-suarez: Removing return_value worked, so I'll commit that. I guess all it needed was side_effect.

tests/unit/test_routes.py
aaron-suarez: I did a quick google search and it looks like there will be a certain amount of research involved for me to learn how to assert that a function was called. Can we turn that into a separate issue and merge this as-is?

app/api/routes.py
wimo7083: Yeah it uses the error level for syslog handling but it also gives some more exception specific details.

tests/unit/test_routes.py
wimo7083: Sounds good to me. The essential mechanics is you create a mock that you provide the ability tell if it's called. It's really helpful for decoupling your tests from the behavior you're testing.

tests/unit/test_routes.py
wimo7083: It's your call, just want to be clear what you are testing.

Issue #109, opened 2019-03-11
Upgrade PyYAML


aaron-suarez:
PyYAML has a new version, we should update it to 4.2b1 or later in requirements.txt

```
pyyaml>=4.2b1
```

phongvdoan:
I will do this to help me learn git

aaron-suarez:
Closed in #118

PR #118, opened 2019-03-28
Update pyyaml requirements to 4.2b1 or newer

phongvdoan:
Issue #109

### DIFF
--- a/app/cli.py
+++ b/app/cli.py
@@ -10,7 +10,7 @@ from .models import Category, Language, Resource
 def import_resources(db):
     # Step 1: Get data
     with open('resources.yml', encoding='utf-8') as f:
-        data = yaml.load(f)
+        data = yaml.full_load(f)
 
     # Step 2: Uniquify resources
     unique_resources = remove_duplicates(data)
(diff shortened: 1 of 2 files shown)


aaron-suarez:
@phongvdoan I just got a chance to try it out and run it. When you run `make setup` now, it lets us know that our use of `yaml.load` is deprecated. The error reads as follows:

```
/src/app/cli.py:13: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  data = yaml.load(f)
```

Since we're updating the dependency in this PR, can we go ahead and update our use of that function before I merge it? You should be able to find correct usage in the docs for PyYAML, if you have trouble let me know and I'll point you in the right direction.

Thanks so much for your contribution!

phongvdoan:
Ok, I will work on it today/tomorrow.

phongvdoan:
From the docs, I got PyYAML version 5.1, you will get a warning, but the function will still work, but I can disable the warning.

phongvdoan:
Or I can specify the Loader to be either BaseLoader, SafeLoader, or FullLoader. FullLoader seems like the default for 5.1, but if we are loading untrusted input then SafeLoader would be recommended..

Jeez I am learning a lot.

phongvdoan:
It is done!

aaron-suarez:
Looks great, thanks for adding that change as well. I'm also glad you learned a lot!

Issue #112, opened 2019-03-12
Method Not Allowed still returning HTML


aaron-suarez:
We need to ensure that we only return JSON responses. In the case of `405 Method Not Allowed`, we're still returning HTML. We need to add a handler in `app/errors/handlers.py` for handling `405` errors, and it should return a standardized response with JSON describing that the method is not allowed.

To avoid merge conflicts and double work, please wait until #110 is merged to start on this issue.

phongvdoan:
when adding handler in handlers.py, does the standardize_response function need to be updated with the 405 error in utils.py?

aaron-suarez:
Yes, but also, a handler has to be added to `app/errors/handlers.py` to catch the 405 error and then call `standardize_response` with a 405 status code

aaron-suarez:
closed in #119

PR #119, opened 2019-04-02
Issue #112 method not allowed

phongvdoan:
Added handler for 405 in app/error/handlers.py and added 405 in err_map


aaron-suarez:
@phongvdoan I forgot to have you add a test around this. Would you mind opening another PR so that has a unit test making a request to a method that's not allowed and verifies that a JSON response is returned?

You can check out an example of me doing something similar with 400 responses here https://github.com/OperationCode/resources_api/pull/123/files#diff-51f932e4d7e5be54d3b9078cf8db7be2R272

PR #120, opened 2019-04-04
Disable strict slashes

aaron-suarez:
I've gotten caught up by this so many times that I'm rather annoyed by it. I fully understand the reasons for having strict slashes, but I think that from a usability perspective, I'd rather the API not return a 404 just because someone put in a slash at the end.


app/__init__.py
```diff
@@ -15,6 +15,7 @@
 def create_app(config_class=Config):
     app = Flask(__name__, static_folder=None)
     app.config.from_object(config_class)
+    app.url_map.strict_slashes = False
```
wimo7083: Awesome. I like this one.

Issue #116, opened 2019-03-25
Testing should be decoupled from specific response strings


aaron-suarez:
Per this comment https://github.com/OperationCode/resources_api/pull/111#discussion_r266947365

Some tests basically test the exact string returned from the server rather than the fact that the exception handler is called. This can be problematic if we ever need to change those strings and the exception is still called fine (meaning the test should pass). An example of this is here:
https://github.com/OperationCode/resources_api/blob/10c20bf8d5df5ee4b387eb79da6bf75ae90d9b83/tests/unit/test_routes.py#L299

There are several instances of this throughout the tests.

RanadeepPolavarapu:
Hello, I could tackle this.

PR #122, opened 2019-04-04
Add basic validation to create resource endpoint

### LINKED ISSUE #113: Failure reason not clear on create resource
When you attempt to create a resource via `POST /api/v1/resources`, if you are missing required fields, you get a `400 Bad Request`, but it doesn't tell you why the request was bad. The API should tell you what fields you're missing. Also, if the only problem is that you're missing a required field, it should probably return a 422 instead of a 400.

aaron-suarez:
Fixes #113 

Now, if you're missing a required parameter (currently, `name`, `url`, and `category`), you will get a response similar to the following:
```json
{
    "apiVersion": "1.0",
    "data": null,
    "errors": {
        "missing_params": [
            "name",
            "url",
            "category"
        ],
        "type": "validation"
    },
    "status_code": 422
}
```

### DIFF
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -25,6 +25,9 @@ def resources():
 @bp.route('/resources', methods=['POST'], endpoint='create_resource')
 @authenticate
 def post_resources():
+    validation_errors = validate_resource(request.get_json())
+    if validation_errors:
+        return standardize_response(payload=validation_errors, status_code=422)
     return create_resource(request.get_json(), db)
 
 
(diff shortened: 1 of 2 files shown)


tests/unit/test_routes.py
```diff
@@ -238,6 +238,17 @@ def test_create_resource(module_client, module_db, fake_auth_from_oc):
...
+    # Invalid Resource Path
+    response = client.post('/api/v1/resources',
+        headers = {'x-apikey': apikey}
+    )
+    assert (response.status_code == 422)
+    response = client.post('/api/v1/resources',
+        json = dict(notes="Missing Required fields"),
+        headers = {'x-apikey': apikey}
+    )
+    assert (response.status_code == 422)
```
aaron-suarez: I would like to validate the message and such, but I decided against doing that here because of #116

app/api/routes.py
```diff
@@ -323,3 +326,27 @@ def create_new_apikey(email):
...
+    validation_errors = {'errors': {'type': 'validation'}}
+    if not json:
+        message = "A JSON body is required to use this endpoint, but none was given"
+        validation_errors['errors']['message'] = message
+        return validation_errors
+
+    validation_errors['errors']['missing_params'] = []
+
+    required = []
+    for column in Resource.__table__.columns:
```
wimo7083: What's an example of this?

aaron-suarez: Basically, the `for` loop goes through each column for the `resources` table and checks for all the ones that are `nullable=False`. There are 4 (but one of them is the `id`, so I explicitly don't check for that). The remaining 3 are "name", "category", and "url". You can't have a resource that doesn't have all 3 of those attributes.

Does that answer your question?

aaron-suarez:
Do you think that `response.errors` should be an array of error objects even if it only contains a single error? That might make more sense than either limiting ourselves to a single error, or having a different type depending on whether there is one or multiple errors encountered.... But like, if creating a resource fails based on validation, it won't go any further than that. Is there ever a situation in which there are multiple errors? If not, perhaps we should change that key to `error` instead of `errors`

app/api/routes.py
aaron-suarez: After #133 gets merged, there will be 4 required fields: `name`, `category`, `url`, and now `paid`.

PR #123, opened 2019-04-04
Add handler for 400 bad request

### LINKED ISSUE #121: 400 response for malformed JSON returns HTML
If you send a malformed JSON request, such as the following, which is missing a comma:

```json
{
    "name": "test"
    "category": "test"
}
```
The API responds with an HTML response such as:
```html
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
<title>400 Bad Request</title>
<h1>Bad Request</h1>
<p>Failed to decode JSON object: Expecting ',' delimiter: line 3 column 2 (char 19)</p>
```

We should be catching errors like this and returning a `standardize_response` for them.

aaron-suarez:
Fixes #121

### DIFF
--- a/app/errors/handlers.py
+++ b/app/errors/handlers.py
@@ -4,6 +4,15 @@ from app.utils import standardize_response
 
 
 # Error Handlers
+@bp.app_errorhandler(400)
+def bad_request(e):
+    error = dict(
+        code="bad-request",
+        message=str(e)
+    )
+    return standardize_response(payload=dict(errors=[error]), status_code=400)
+
+
 @bp.app_errorhandler(404)
 def page_not_found(e):
     return standardize_response(status_code=404)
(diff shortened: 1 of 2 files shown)


app/errors/handlers.py
```diff
@@ -4,6 +4,15 @@
 
 
 # Error Handlers
+@bp.app_errorhandler(400)
+def bad_request(e):
+    error = dict(
+        code="bad-request",
+        message=str(e)
```
aaron-suarez: Normally I advocate against passing along server error messages directly, but in the case of `400 - Bad Request`, I feel like the user should know why their request is bad. This seemed the simplest way to keep that usability, and I believe the security risk is rather low. But we need to make sure no one cargo cults this pattern for a more sensitive error

PR #124, opened 2019-04-05
Feature/add pagination details

Qlwentt:
This feature adds the following json blob any paginated response:
```
"page": <page_number>,
"number_of_pages": <total_number_of_pages>,
"records_per_page": <records_per_page>,
```
To implement this, I decided to change `items()` in `Paginator` to return the full paginated data object and renamed it `paginated_data()` because to get the pagination details we need the object returned from `query.paginate` (already called in `items()`) and I didn't want to make two calls. To get items, we now do `paginated_data().items`

I then added the pagination details in the `standardize_response` method when paginated_data is present (so only for paginated responses). I also updated the controller actions that use pagination to be compatible with my changes.

Finally, I ensured pagination details were getting returned in the test`test_paginator`. Because of the change to `items()` I also had to update `fake_items_error(mocker)` in `conftest.py` to be `paginated_data_error(mocker)`.

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -14,11 +14,13 @@ class Paginator:
         if self.page_size > configuration.max_page_size:
             self.page_size = configuration.max_page_size
 
-    def items(self, query):
-        return query.paginate(self.page, self.page_size, False).items
 
+    def paginated_data(self, query):
+        data = query.paginate(self.page, self.page_size, False)
+        setattr(data, "per_page", self.configuration.per_page)
+        return data
 
-def standardize_response(payload=None, status_code=200):
+def standardize_response(payload=None, status_code=200, paginated_data=None):
     """Response helper
     This simplifies the response creation process by providing an internally
     defined mapping of status codes to messages for errors. It also knows when
@@ -71,6 +73,11 @@ def standardize_response(payload=None, status_code=200):
         resp["status"] = "error"
     else:
         resp["data"] = data
+
+    if paginated_data:
+        resp["page"] = paginated_data.page
+        resp["number_of_pages"] = paginated_data.pages
+        resp["records_per_page"] = paginated_data.per_page
 
     return jsonify(resp), resp["status_code"], {'Content-Type': 'application/json'}
 
--- a/app/utils.py
+++ b/app/utils.py
@@ -19,8 +19,17 @@ class Paginator:
         setattr(data, "per_page", self.configuration.per_page)
         return data
 
+    def pagination_details(self, paginated_data):
+        return {
+            "pagination_details": {
+                "page": paginated_data.page,
+                "number_of_pages": paginated_data.pages,
+                "records_per_page": paginated_data.per_page
+            }
+        }
 
-def standardize_response(payload=None, status_code=200, paginated_data=None):
+
+def standardize_response(payload={}, status_code=200):
     """Response helper
     This simplifies the response creation process by providing an internally
     defined mapping of status codes to messages for errors. It also knows when
@@ -34,10 +43,9 @@ def standardize_response(payload=None, status_code=200, paginated_data=None):
     status_code -- a valid HTTP status code. For errors it defaults to 500,
     for 'ok' it defaults to 200
     """
-    if not payload:
-        payload = {}
     data = payload.get("data")
     errors = payload.get("errors")
+    pagination_details = payload.get("pagination_details")
     resp = dict(
         apiVersion=API_VERSION,
         status="ok",
@@ -74,10 +82,8 @@ def standardize_response(payload=None, status_code=200, paginated_data=None):
     else:
         resp["data"] = data
 
-    if paginated_data:
-        resp["page"] = paginated_data.page
-        resp["number_of_pages"] = paginated_data.pages
-        resp["records_per_page"] = paginated_data.per_page
+    if pagination_details:
+        resp.update(pagination_details)
 
     return jsonify(resp), resp["status_code"], {'Content-Type': 'application/json'}
 
--- a/app/utils.py
+++ b/app/utils.py
@@ -19,7 +19,7 @@ class Paginator:
         if self.page > data.pages:
             return None
 
-        setattr(data, "per_page", self.configuration.per_page)
+        setattr(data, "per_page", self.page_size)
         return data
 
     def pagination_details(self, paginated_data):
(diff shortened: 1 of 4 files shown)


app/utils.py
```diff
@@ -14,11 +14,13 @@ def __init__(self, configuration, request):
...
-    def items(self, query):
-        return query.paginate(self.page, self.page_size, False).items
+    def paginated_data(self, query):
+        data = query.paginate(self.page, self.page_size, False)
+        setattr(data, "per_page", self.configuration.per_page)
+        return data
 
 
-def standardize_response(payload=None, status_code=200):
+def standardize_response(payload=None, status_code=200, paginated_data=None):
```
aaron-suarez: The only thing I'm on the fence about is having `paginated_data` as its own kwarg.  This function used to take like 5 different arguments and it was very unweildy, and @lord-cant-even came in and fixed that by making it just `payload` and `status_code`. The intent of `payload` was that it could have all of the data in that dictionary and not have a whole bunch of arguments to the method.

I'm willing to concede on this if you have an argument for why an additional kwarg is better, but if it's literally all the same to you, I'd prefer to keep this data in `payload` rather than add any additional arguments to `standardize_response`.

aaron-suarez: And if you do decide to change it, can you also change the default for `payload` to be an empty dictionary and delete these lines?
https://github.com/OperationCode/resources_api/blob/89116af17ffc48b17bbd459ec9eb920f3d3b540a/app/utils.py#L35-L36

I didn't realize until just now that these are unnecessary if we just make the default for `payload` an empty dictionary

Qlwentt: Thanks for the background for `standardize_response` !
The reason I added the pagination details there is because I was trying to find a way to add the pagination details in one place so they could get added for all of the actions that use pagination.

But maybe it makes more sense to add a `pagination_details` function in `utils.py`. We'd pass in `paginated_data` and it would return
```
{
    "page" :  paginated_data.page
    "number_of_pages" : paginated_data.pages
    "records_per_page": paginated_data.per_page
}
```

then I could call that function in all of the actions that use pagination. and then merge that dictionary with the rest of the payload. That way when people add more actions that use pagination they don't have to include each pagination detail individually, they just know they need to call `pagination_details` and add whatever is returned.

What do you think?

Also, sure, I can change the the default for payload to be an empty dict and delete lines 35-36 👍

aaron-suarez: I'm not going to lie, I'm not entirely sure what that would look like but it sounds good to me. Anything that merges that data with the payload satisfies my distaste with the current situation, so I'm down to check it out. Please make those changes and then I'll review it again and we'll iterate til we get it (but I'm pretty sure this next iteration will be fine if I understand your suggestion correctly). Thanks for your contribution 😄

Qlwentt:
I added a check for pagination details in `test_categories` and `test_languages` because I realized that its possible for `test_paginator` to pass but the pagination details to still be missing in the language and categories get responses.

aaron-suarez:
Thanks for updating the tests, good call!

I'm mobile right now, so I can't run it, but I'm sure it's fine. I'm going to approve it for now and merge it once I verify. This is great, thanks 😁

aaron-suarez:
After playing around with it a little bit, I'm wondering if it would be helpful to add a `count:` field in the pagination details that specifies how many records were returned in this particular `GET` request? If you're consuming it with a JavaScript client, it would be trivial to just call `response.data.length` or whatever, but when you view it via Postman or a browser, it might be helpful to have the count right there.

aaron-suarez:
Not sure if this has anything to do with your code or if the error existed previously, but when I try to access a page that is too high (example: `number_of_pages: 94` but I requested `GET http://127.0.0.1:8000/api/v1/resources?page=95`) it returns a 500 error:
```json
{
    "apiVersion": "1.0",
    "data": null,
    "errors": [
        {
            "code": "something-went-wrong"
        }
    ],
    "number_of_pages": 94,
    "page": 95,
    "records_per_page": 20,
    "status": "error",
    "status_code": 500
}
```
When the page number is higher than the page count, it should return a 404, not a 500.

We have a test for this type of thing on the `votes` route, could we add one on the `GET /api/v1/resources`, `languages`, and `categories` routes too?
https://github.com/OperationCode/resources_api/blob/cbc9d6dd03c31a42aad9f615fc611b7bd29188f5/tests/unit/test_routes.py#L178-L182

Qlwentt:
Just double checking, you mean how many total records are returned across all the pages of results right? This would be the `total` attribute on the Pagination object, so I could easily add it.

Also while we are at it, are there any other details from the pagination object that we should include (like `has_next`, `has_previous`, `next_num`, `prev_num`)? These docs have everything that's available for the object: http://flask-sqlalchemy.pocoo.org/2.3/api/#flask_sqlalchemy.Pagination

Qlwentt:
ok I can definitely fix the page too high error and make sure its a 404 instead of a 500. 👍

aaron-suarez:
Regarding the other attributes such as `has_next`, I think that could be useful. We won't really know what we need for the MVP until we get to work on the front end, but I'll defer to you on what you want to add, I definitely see the value. 

As for my original statement, I realize now that I just meant how many records are on the last page since there may be less than whatever is in the `per_page` attribute. But if you return the total count across all pages, it would be a simple subtraction problem on the front end, so that would probably suffice.

aaron-suarez:
Thanks for making that change! Again, I'm not sure if this is due to the new code or if this was a previously existing bug, but it would be nice to fix it in this PR as well since the point is to fix all the pagination issues.

`configs.py` specifies a `max_page_size` of 100:
https://github.com/OperationCode/resources_api/blob/cbc9d6dd03c31a42aad9f615fc611b7bd29188f5/configs.py#L9

However, when I do `GET http://127.0.0.1:8000/api/v1/resources?page_size=25`, I only get 20 records per page. I thought perhaps I had the name of the query argument wrong, so I tried `GET http://127.0.0.1:8000/api/v1/resources?per_page=25` and that still returned 20 records as well. The paginator should return up to 100 records (or up to whatever is in that `max_page_size` class variable for the paginator), so this behavior is incorrect.

aaron-suarez:
Oh and if we could get a test around that as well, that would be awesome 😄

Qlwentt:
the response actually had the correct number of records in it, but I was returning the wrong value for `number_of_records`. I fixed that.

`max_page_size` refers to the maximum number of records, so even if the user requests 101 records (`page_size=101`) per page, they will only get 100 because `max_page_size` is 100.

it's now working as intended and I added tests to verify. Thanks for catching that!

aaron-suarez:
Ok, looks great, thanks!

Issue #76, opened 2018-12-24
Need route for times a resource is clicked


aaron-suarez:
When the front end is created, we're going to track the popularity of a resource by tracking how many times people click the link to open that resource. Eventually, we're also going to track who specifically clicks on the resource so that we don't increment `times_clicked` if they have already clicked the resource in the past, but that is beyond the scope of this issue.

The route should be `PUT api/v1/resources/<id>/times_clicked` and it should return the updated resource with the `times_clicked` value incremented. Please write a test for this in `test_routes.py`.

aaron-suarez:
Upon further thinking, I'm pretty sure I'd rather the route be `PUT api/v1/resources/<id>/click`, because click is the action.

aaron-suarez:
Closed via #125

PR #125, opened 2019-04-06
Feature/resources click route

Qlwentt:
For Issue #76 

Adds route: `PUT api/v1/resources/<id>/click` 
increments `times_clicked` attribute for Resource

Adds a test to validate the route is working (also tested locally)

### DIFF
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -49,6 +49,11 @@ def downvote(id):
     return update_votes(id, 'downvotes')
 
 
+@bp.route('/resources/<int:id>/click', methods=['PUT'])
+def update_resource_click(id):
+    return add_click(id)
+
+
 @bp.route('/languages', methods=['GET'])
 def languages():
     return get_languages()
@@ -246,6 +251,34 @@ def update_votes(id, vote_direction):
 
     initial_count = getattr(resource, vote_direction)
     setattr(resource, vote_direction, initial_count+1)
+    db.session.commit()
+
+    return standardize_response(payload=dict(data=resource.serialize))
+
+
+def add_click(id):
+    try:
+        resource = Resource.query.get(id)
+
+        if not resource:
+            return redirect('/404')
+
+    except MultipleResultsFound as e:
+        print_tb(e.__traceback__)
+        logger.exception(e)
+
+    except NoResultFound as e:
+        print_tb(e.__traceback__)
+        logger.exception(e)
+        return redirect('/404')
+
+    except Exception as e:
+        print_tb(e.__traceback__)
+        logger.exception(e)
+        return standardize_response(status_code=500)
+
+    initial_count = getattr(resource, 'times_clicked')
+    setattr(resource, 'times_clicked', initial_count + 1)
     db.session.commit()
 
     return standardize_response(payload=dict(data=resource.serialize))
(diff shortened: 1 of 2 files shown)


tests/unit/test_routes.py
```diff
@@ -182,6 +182,31 @@ def test_update_votes(module_client, module_db):
...
+    id = 1
+    data = client.get(f"api/v1/resources/{id}").json['data']
+    response = client.put(f"/api/v1/resources/{id}/click", follow_redirects=True)
+    initial_click_count = data.get(f"times_clicked")
+
+    assert (response.status_code == 200)
+    assert (response.json['data'].get(f"times_clicked") == initial_click_count + 1)
+
+    # Check clicking on an invalid resource
+    id = 'pancakes'
```
aaron-suarez: I appreciate the negative unit tests!

app/api/routes.py
```diff
@@ -251,6 +256,34 @@ def update_votes(id, vote_direction):
...
+def add_click(id):
+    try:
+        resource = Resource.query.get(id)
+
+        if not resource:
+            return redirect('/404')
+
+    except MultipleResultsFound as e:
+        print_tb(e.__traceback__)
+        logger.exception(e)
```
aaron-suarez: I've never been able to actually create a `MultipleResultsFound` condition. But let's suppose that it happened, what would this function return?

If this was basically copy-pasted from other parts of the routes file, we should probably open an issue that these error conditions should return early with some sort of error to the client. Maybe a `500` would do since we're logging the error?

Not a blocker for the PR, but something we should think about

Qlwentt: Ah yes, I did copy and paste this from another route (upvotes).

Yeah I don’t know how it would be possible to find multiple resources with the same id. That would be impossible database-wise. Maybe we should remove this condition altogether?

aaron-suarez: Considering that I've never run into it and I can't produce it for unit tests, it would be good for test coverage purposes to go ahead and just delete it on all routes. I agree that it makes no sense that it should ever return multiple results when we have `id` as the primary key and specified as unique in the model. If this ever happens, I hope it just crashes everything and we are forced to clean up the data lol.

Please open an issue documenting this or just a PR to remove all of them. If we run into some strange issue in prod, I'll take on the task/responsibility of troubleshooting and fixing it. Thanks for the second set of eyes on this.

Qlwentt: Sounds good! I’ll create an issue for this for documentation sake and prolly submit the PR tomorrow.

aaron-suarez: Awesome, thanks! You've been such a great help to pushing this over the finish line, I very much appreciate that. This is the most momentum we've gotten for the project in months and it couldn't have come at a better time

Qlwentt: My pleasure! I’m happy to contribute! This is the first time I’ve contributed to open source and it’s been an awesome experience so far :D

aaron-suarez: That makes me feel good. At my day job, I'm on the "Developer Experience" team, maintaining our API and open source client libraries that our customers use to interact with it. Our number 1 mission is to make contributing to and using our code a delightful and seamless experience, and I'm glad that I can bring that same type of delight to contributors to Operation Code because it's an organization I care a lot about. 

Anyway, this project is just about ready to go live and it's my first open source project that I personally came up with and saw from design to production, so this is huge for me. :excited:

Issue #69, opened 2018-11-25
Better error handling of out of bounds GET resources


aaron-suarez:
As of the latest commit, if you try to GET a resource with an id that's higher than the highest in the database, you get an error that looks like this:

```json
{
    "apiVersion": "1.0",
    "errors": [
        {
            "code": "something-went-wrong"
        }
    ],
    "status": "ok"
}
```

I put this error in here as a "catch all" for errors that we aren't specifically catching. We need to specifically catch the out-of-bounds scenario I described.

Acceptance criteria:
When doing `GET /api/v1/resources/10000000`, a 404 error should be returned.

aaron-suarez:
Update: As of the latest commit, it now returns something like this:
```json
{
    "apiVersion": "1.0",
    "data": {},
    "status": "ok"
}
```

But it should still be a `404` not found instead

aaron-suarez:
Fixed in #129

PR #129, opened 2019-04-08
Fix the redirect when resource is not found

aaron-suarez:
I was getting a 500 error when I would do `GET http://127.0.0.1:8000/api/v1/resources/9999`. This fixed that to make it a 404 instead.


aaron-suarez:
Fixes #69 (finally)

PR #133, opened 2019-04-14
Add filter for paid/free resources

### LINKED ISSUE #130: Issues with paid resources
1) Users should be able to filter resources by whether the `paid` attribute is true or false. This will be useful when searching specifically for free resources. This might be suitable to implement on the client side, but I feel like having less data will be preferable for mobile clients. 

2) Each resource should have an attribute called "paid" which should be true or false. I do not believe that any resources in the `resources.yml` file are missing this attribute, but I don't think we check for it when creating a new resource. We need to validate that this attribute exists in the Resources model (non-nullable).

aaron-suarez:
Fixes #130 

This filter works for me locally, but for some reason when the query occurs in the tests it returns no resources and then the response redirects to `404` (which would be correct if there were no resources). I'm confused why this is happening since I'm pretty sure it's the same seed data in `module_db` as in my local db. Help fixing the test would be greatly appreciated!

### DIFF
--- a/tests/unit/test_routes.py
+++ b/tests/unit/test_routes.py
@@ -137,6 +137,30 @@ def test_paginator(module_client, module_db):
     assert (response['has_prev'] is not None)
 
 
+def test_paid_filter(module_client, module_db):
+    client = module_client
+
+    total_resources = client.get('api/v1/resources').json['total_count']
+
+    # Filter by paid
+    response = client.get('api/v1/resources?paid=false')
+
+    total_free_resources = response.json['total_count']
+
+    assert all([res.get('paid') == False for res in response.json['data']])
+
+    response = client.get('api/v1/resources?paid=true')
+
+    total_paid_resources = response.json['total_count']
+
+    assert all([res.get('paid') == True for res in response.json['data']])
+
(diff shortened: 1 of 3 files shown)


tests/unit/test_routes.py
```diff
@@ -137,6 +137,33 @@ def test_paginator(module_client, module_db):
...
+    total_free_resources = response.json['total_count']
+
+    for resource in response.json['data']:
+        assert (resource.get('paid') == False)
+
+    response = client.get('api/v1/resources?paid=true')
+
+    total_paid_resources = response.json['total_count']
+
+    for resource in response.json['data']:
```
wimo7083: assert is an expression
also you can do this:

```
assert all([res.get('paid') == False for res in resources])
```

tests/unit/test_routes.py
```diff
@@ -137,6 +137,33 @@ def test_paginator(module_client, module_db):
...
+        assert (resource.get('paid') == True)
+
+    # Check that the number of resources appear correct
+    assert (total_resources > 0)
```
wimo7083: If our current seed data has at least one this works.

PR #138, opened 2019-04-20
Remove unnecessary environment variables

aaron-suarez:
When we originally moved to docker, we weren't sure which environment variables to use or what they should be. As soon as we got it working, we just left everything how it was. Now it's time to clean this up so we can go to production. This passes all the tests and appears to work correctly on my machine, if it passes CI, I have full confidence in this change.


.env
```diff
@@ -2,15 +2,7 @@ FLASK_APP=run.py
 SQLALCHEMY_DATABASE_URI=postgresql://user_name:change_password@resources-psql:5432/resources-psql
```
wimo7083: Can you have the application build SQLALCHEMY_DATABASE_URI from the provided variables?

.env
```diff
@@ -2,15 +2,7 @@ FLASK_APP=run.py
...
-DB_HOST=postgresql
-DB_PORT=5432
-DB_DATABASE=resources-psql
-DB_USERNAME=user_name
-DB_CONNECTION=pgsql
-USER=user_name
-PASSWORD=change_password
-DB=resources-psql
+POSTGRES_DB=resources-psql
 FLASK_SKIP_DOTENV=1
```
wimo7083: Can FLASK_SKIP_DOTENV be set from a config instead of environment variable?

docker-compose.yml
```diff
@@ -22,7 +22,7 @@ services:
     environment:
     - POSTGRES_USER=user_name
```
wimo7083: POSTGRES_USER and POSTGRES_PASSWORD should be environment variables also.

.env
aaron-suarez: http://flask.pocoo.org/docs/1.0/cli/#disable-dotenv

wimo7083: So no it can't, and no we can't get this to work by adjusting the dockerfile to remove the `.env` and by adding the files other than `.env`

Attempting to set these application level environments in the dockerfile by using the `ENV` command.

aaron-suarez: Set it in the Dockerfile

PR #142, opened 2019-04-27
Improve test coverage

Digbigpig:
Addresses #79 
Increased code coverage by excluding import resources as it isn't needed in
production.
Removed blocks catching NoResultsFound on get() queries.

### DIFF
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -2,7 +2,6 @@ from traceback import print_tb
 
 from flask import request, redirect
 from sqlalchemy import or_, func
-from sqlalchemy.orm.exc import NoResultFound
 
 from app.api import bp
 from app.api.auth import is_user_oc_member, authenticate
@@ -99,14 +98,7 @@ def apikey():
 
 # Helpers
 def get_resource(id):
-    resource = None
-    try:
-        resource = Resource.query.get(id)
-
-    except NoResultFound as e:
-        print_tb(e.__traceback__)
-        logger.exception(e)
-        return redirect('/404')
+    resource = Resource.query.get(id)
 
     if resource:
         return standardize_response(payload=dict(data=(resource.serialize)))
@@ -251,11 +243,6 @@ def update_votes(id, vote_direction):
         if not resource:
             return redirect('/404')
 
-    except NoResultFound as e:
-        print_tb(e.__traceback__)
-        logger.exception(e)
-        return redirect('/404')
-
     except Exception as e:
         print_tb(e.__traceback__)
         logger.exception(e)
@@ -274,11 +261,6 @@ def add_click(id):
 
         if not resource:
             return redirect('/404')
-
-    except NoResultFound as e:
-        print_tb(e.__traceback__)
-        logger.exception(e)
-        return redirect('/404')
 
     except Exception as e:
         print_tb(e.__traceback__)
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -1,5 +1,3 @@
-from traceback import print_tb
-
 from flask import request, redirect
 from sqlalchemy import or_, func
 
@@ -237,16 +235,11 @@ def get_attributes(json):
 
 
 def update_votes(id, vote_direction):
-    try:
-        resource = Resource.query.get(id)
 
-        if not resource:
-            return redirect('/404')
+    resource = Resource.query.get(id)
 
-    except Exception as e:
-        print_tb(e.__traceback__)
-        logger.exception(e)
-        return standardize_response(status_code=500)
+    if not resource:
+        return redirect('/404')
 
     initial_count = getattr(resource, vote_direction)
     setattr(resource, vote_direction, initial_count+1)
@@ -256,16 +249,10 @@ def update_votes(id, vote_direction):
 
 
 def add_click(id):
-    try:
-        resource = Resource.query.get(id)
+    resource = Resource.query.get(id)
 
-        if not resource:
-            return redirect('/404')
-
-    except Exception as e:
-        print_tb(e.__traceback__)
-        logger.exception(e)
-        return standardize_response(status_code=500)
+    if not resource:
+        return redirect('/404')
 
     initial_count = getattr(resource, 'times_clicked')
     setattr(resource, 'times_clicked', initial_count + 1)
@@ -275,7 +262,6 @@ def add_click(id):
 
 
 def set_resource(id, json, db):
-    resource = None
     resource = Resource.query.get(id)
 
     if not resource:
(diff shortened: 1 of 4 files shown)


app/api/routes.py
```diff
@@ -99,14 +98,7 @@ def apikey():
...
 def get_resource(id):
-    resource = None
-    try:
-        resource = Resource.query.get(id)
-
-    except NoResultFound as e:
-        print_tb(e.__traceback__)
-        logger.exception(e)
-        return redirect('/404')
+    resource = Resource.query.get(id)
```
Digbigpig: In all the places that I removed the NoResultsFound Exception, it's because get() cannot throw noresultsfound. The .one() function can though.
[Heres the link to the docs](https://docs.sqlalchemy.org/en/13/orm/query.html#sqlalchemy.orm.query.Query.get)
Let me know if I'm crazy for thinking that.

I want to add a catch for ObjectDeletedError but I'm not sure how to even throw the error yet. 
If someone knows how to properly do that, let me know and I can make that change. Tht would improve the coverage quite a bit too

app/cli.py
```diff
@@ -7,7 +7,7 @@
 from .models import Category, Language, Resource
 
 
-def import_resources(db):
+def import_resources(db):   # pragma: no cover
```
Digbigpig: This is how you exclude functions for the test coverage

app/api/routes.py
wimo7083: Can you explain your issue with ObjectDeletedError?

We currently don't have route to delete, and per @aaron-suarez we wouldn't physically delete it but tag the row as non-available and then update the query/ serializer to respect that.

app/api/routes.py
```diff
@@ -99,14 +98,7 @@ def apikey():
 
 # Helpers
 def get_resource(id):
-    resource = None
-    try:
```
wimo7083: I think this is fine, and you're right about the results. This was probably due to copy pasting or occurred as we changed how we used the query api.

tests/unit/test_routes.py
```diff
@@ -83,11 +83,11 @@ def test_single_resource_out_of_bounds(module_client, module_db):
 
     too_low = 0
     too_high = 9999
-    response = client.get(f"api/v1/resources{too_low}")
+    response = client.get(f"api/v1/resources/{too_low}", follow_redirects=True)
```
wimo7083: The PR doesn't mention this, can you put something about why you felt this was needed?

tests/unit/test_routes.py
```diff
@@ -83,11 +83,11 @@ def test_single_resource_out_of_bounds(module_client, module_db):
...
 
     assert (response.status_code == 404)
 
-    response = client.get(f"api/v1/resources{too_high}")
+    response = client.get(f"api/v1/resources/{too_high}", follow_redirects=True)
```
wimo7083: The PR doesn't mention this, can you put something about why you felt this was needed?

app/cli.py
wimo7083: LG2M

app/cli.py
```diff
@@ -115,7 +115,7 @@ def create_resource(resource, db):
         print('exception', e)
 
 
-def update_resource(resource, existing_resource):
+def update_resource(resource, existing_resource):   # pragma: no cover
```
wimo7083: Did you want to do this to all the CLI functions or did you find non-cli usages of them?

wimo7083:
Looking good, just want to get some answers for the questions and then we can merge.

app/api/routes.py
Digbigpig: Well, then we can probably just remove the try blocks all together on those functions. Then we wont need to test those non existent exceptions.

My only "problem" with it was that it is the only exception that the get() method can throw, so it would be the only exception we would need to catch.

app/cli.py
Digbigpig: I only checked the ones that weren't already covered. 
I'll go back over it because there is one more I could probably make exempt.

tests/unit/test_routes.py
Digbigpig: My second commit covers it. Basically these two tests were false positives because the url in client.get was wrong (missing a /) and returning None due to a bad url, not an empty query.

We have to use the follow_redirect because the code redirects to the 404 page.

app/api/routes.py
aaron-suarez: Oh that makes sense. But yes, at this time we do not have a way to delete anything. If we do add a route to delete resources/languages/categories, then we'll need to make sure we're catching that. But even so, I think we would just mark as deleted in the db and it would just filter "deleted" results out rather than throwing that error. Either way, we'll address it if and when it comes up. For now, I don't think we need to worry about it

tests/unit/test_routes.py
aaron-suarez: Oh interesting. Yes, you'll need to follow redirects to get the 404 because if you don't you get a 302. I've encountered that before elsewhere in these tests. Good catch!

aaron-suarez:
I believe `register` is also only used for these same click commands. Can we add an exclusion for that as well?
https://github.com/OperationCode/resources_api/blob/8d028dfaa5f1358af7d858f7df5c5005f60e34f9/app/cli.py#L127-L145

aaron-suarez:
Since we're improving test coverage, can we go ahead and add a test around the `405` error handler?
https://github.com/OperationCode/resources_api/blob/8d028dfaa5f1358af7d858f7df5c5005f60e34f9/app/errors/handlers.py#L21-L23

I don't anticipate us ever using `PATCH` for this API, so that would probably be a good one to use for the test. Make sure that if you send `PATCH /api/v1/resources` that it returns a JSON response with a 405 status code.

aaron-suarez:
If we're comfortable removing the try/except blocks from `get_resource()`, we should also go ahead and remove them from `update_votes()` and `add_click()` because right now they also wrap `resource = Resource.query.get(id)`
https://github.com/OperationCode/resources_api/blob/a0ecc90505a871702628330cf1b193470dcefd60/app/api/routes.py#L239-L268

aaron-suarez:
We can add an exclusion comment to this line as well. I know that it works, we don't need a test covering it
https://github.com/OperationCode/resources_api/blob/a0ecc90505a871702628330cf1b193470dcefd60/app/utils.py#L100

Digbigpig:
The last thing I needed to do is add a test for the error handler. I'm not certain how the code uses the error handlers yet so if you have suggestions, or can point me to other implementations, that will help

aaron-suarez:
@Digbigpig Add this to `test_routes.py`
```python
def test_method_not_allowed_handler(module_client):
    client = module_client

    response = client.patch('api/v1/resources')
    assert (response.status_code == 405)
```

PR #145, opened 2019-04-28
Dockerhub

wimo7083:
Let's try some better bash.

### DIFF
--- a/bin/run.sh
+++ b/bin/run.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+
+# Only process first job in matrix (TRAVIS_JOB_NUMBER ends with ".1")
+if [[ ! $TRAVIS_JOB_NUMBER =~ \.1$ ]]; then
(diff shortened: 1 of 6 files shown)


bin/run.sh
```diff
@@ -15,8 +15,11 @@ fi
 
 # Only process branches listed in DEPLOY_BRANCHES
 BRANCHES_TO_DEPLOY=($DEPLOY_BRANCHES)
-if [[ ! " ${BRANCHES_TO_DEPLOY[@]} " =~ " ${TRAVIS_BRANCH} " ]]; then
+if [[ ! " ${BRANCHES_TO_DEPLOY} " =~ " ${TRAVIS_BRANCH} " ]]; then
   # whatever you want to do when arr contains value
```
aaron-suarez: It's not an array anymore, we should update this comment

wimo7083: Will do.

PR #147, opened 2019-04-28
Use port 5000 instead of 8000

aaron-suarez:
For some strange reason when we switched to Docker, we started using port 8000. I'm not really sure why that was the case, but now in prod we're using the default port when we run flask. I think it's a good idea for local environments to more closely resemble prod, so I made the change in the Dockerfile and elsewhere to use the default port.

### DIFF
--- a/configs.py
+++ b/configs.py
@@ -36,7 +36,6 @@ if not postgres_host:
 class Config:
     SQLALCHEMY_TRACK_MODIFICATIONS = False
     SQLALCHEMY_DATABASE_URI = f"postgresql://{postgres_user}:{postgres_password}@{postgres_host}:5432/{postgres_db}"
-    print(SQLALCHEMY_DATABASE_URI)
 
     # Can pass in changes to defaults, such as PaginatorConfig(per_page=40)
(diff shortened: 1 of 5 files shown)


.env
```diff
@@ -1,6 +1,6 @@
 POSTGRES_USER=user_name
 POSTGRES_PASSWORD=change_password
-POSTGRES_HOST=resources-postgres
+POSTGRES_HOST=resources-psql
```
aaron-suarez: This was broken but I think it went unnoticed because we were using old containers locally. It broke as soon as I did a `make nuke && make fresh-restart`

configs.py
```diff
@@ -36,7 +36,6 @@ def get_sys_exec_root_or_drive():
 class Config:
     SQLALCHEMY_TRACK_MODIFICATIONS = False
     SQLALCHEMY_DATABASE_URI = f"postgresql://{postgres_user}:{postgres_password}@{postgres_host}:5432/{postgres_db}"
-    print(SQLALCHEMY_DATABASE_URI)
```
aaron-suarez: I'm pretty sure we don't need this anymore either

PR #148, opened 2019-04-29
Fix make setup bug

Digbigpig:
Fixes a bug preventing make setup from running.

### DIFF
--- a/.env
+++ b/.env
@@ -1,6 +1,6 @@
 POSTGRES_USER=user_name
 POSTGRES_PASSWORD=change_password
-POSTGRES_HOST=resources-postgres
+POSTGRES_HOST=resources-psql
 POSTGRES_DB=resources-psql
 FLASK_ENV=development
 FLASK_DEBUG=1


.env
```diff
@@ -1,6 +1,6 @@
 POSTGRES_USER=user_name
 POSTGRES_PASSWORD=change_password
-POSTGRES_HOST=resources-postgres
+POSTGRES_HOST=resources-psql
```
wimo7083: I disagree with @aaron-suarez  I think this: https://github.com/OperationCode/resources_api/blob/2e34037463e4eb8be9d7ee13d5ecdbb4a51ae73c/docker-compose.yml#L21

Should be changed to be `resources-postgres` which is how we are doing it in the infrastructure. 

@aaron-suarez if you want it to be `resources-psql` we should do it both here and infrastructure.

aaron-suarez: I don't have a strong opinion on the naming of this, but I do agree that it should be the same in the infrastructure as local. I'll make a PR to change it to `resources-postgres`. My bad

Digbigpig: I tried doing this myself but I'm missing some other steps apparently. Any where else that it needs to be changed that you can think of off the top of your head?

Digbigpig: When I only change those two, the tests all pass but when I try to navigate to the page, it gets a server error. Kind of odd.

PR #150, opened 2019-05-03
Search endpoint

Digbigpig:
I know this will have a lot of changes that need to be made so feedback is appreciated.

First of all, until we get an OC account with algolia, if you want to test this out, you'll need to plug in your own algolia account and index information.
Once you set up your account, just create an index and plug the name into `client.init_index('your index name')`

I'll go through and comment on all the changes I made.

For an example the url: /api/v1/resources/search/dog returns the following json:
```{
  "apiVersion": "1.0", 
  "data": [
    {
      "category": "Front End Dev", 
      "created_at": "2019-04-27 16:22:55", 
      "downvotes": 0, 
      "id": 195, 
      "languages": [
        "HTML"
      ], 
      "last_updated": "", 
      "name": "HTML Dog", 
      "notes": null, 
      "objectID": 195, 
      "paid": false, 
      "times_clicked": 0, 
      "upvotes": 0, 
      "url": "http://www.htmldog.com/"
    }, 
    {
      "category": "Veterans", 
      "created_at": "2019-04-27 16:22:55", 
      "downvotes": 0, 
      "id": 1966, 
      "languages": [], 
      "last_updated": "", 
      "name": "Next Step Service Dogs", 
      "notes": null, 
      "objectID": 1966, 
      "paid": false, 
      "times_clicked": 0, 
      "upvotes": 0, 
      "url": "https://www.nextstepservicedogs.org/"
    }
  ], 
  "number_of_pages": 1, 
  "page": 0, 
  "records_per_page": 20, 
  "status": "ok", 
  "status_code": 200, 
  "total_count": 2
}
```


app/__init__.py
```diff
@@ -11,6 +12,10 @@
 
 API_VERSION = "1.0"
 
+# Connect to Agolia
+client = SearchClient.create('your_project_name', 'your_key')
+index = client.init_index('your index name')
+
```
Digbigpig: I'm not certain how the environment variables are added and we dont have an OC account yet anyways so I left it like this for now.

app/api/routes.py
```diff
@@ -53,6 +53,9 @@ def downvote(id):
 def update_resource_click(id):
     return add_click(id)
 
+@bp.route('/resources/search/<string:term>', methods=['GET'])
```
Digbigpig: Do we want this to be just '/search/ '

app/api/routes.py
```diff
@@ -179,6 +182,57 @@ def get_resources():
...
+            curr_index = ind
+
+            if curr_index['entries'] != query.count():
+                print("ReIndexing Database")
+
+                index.clear_objects()
+                db_list = [u.serialize for u in query.all()]
+                index.save_objects(db_list)
+
+            break
```
Digbigpig: From 192: 204
This block indexes the db to algolia if the database and their index are different sizes. 
I tried to include this code in multiple places where it seemed appropriate, but I was running into import errors trying to get the Query to run. 

I might just have a function that fires it off without checking on startup.

app/api/routes.py
```diff
@@ -179,6 +182,57 @@ def get_resources():
...
+        'page': page,
+    })
+
+    if page >= int(search_results['nbPages']):
+        return redirect('/404')
+
+    # Translate search to normal format.
+    results = []
+    for result in search_results['hits']:
+        results.append(query.get(int(result['objectID'])).serialize)
```
Digbigpig: Algolia returns a lot of meta data on the search, so here we just get the id of each record and return the serialized object from our db.

app/api/routes.py
```diff
@@ -179,6 +182,57 @@ def get_resources():
...
+
+    pagination_details = {
+                "pagination_details": {
+                    "page": search_results['page'],
+                    "number_of_pages": search_results['nbPages'],
+                    "records_per_page": search_results['hitsPerPage'],
+                    "total_count": search_results['nbHits'],
+                }
+        }
+    return standardize_response(payload=dict(data=results, **pagination_details))
```
Digbigpig: Since Algolia does its own pagnation, we dont need to use the pagnator to 
1. Pagnate the data
2. Generate the pagnation details.

There's probably a better way to handle this.

Digbigpig:
I'm assuming we want to add filtering to the search too which I haven't started on yet. 
`Category, language, paid` filters seem like the obvious ones.

app/api/routes.py
Digbigpig: Also, our paginator starts at 1 and Algolias starts at 0. I dont know if that is something we should care about

app/models.py
```diff
@@ -50,6 +50,7 @@ def serialize(self):
 
         return {
             'id': self.id,
+            'objectID': self.id,
```
Digbigpig: Algolia uses this property name on indexing.

app/__init__.py
kylemh: Likely through Travis on prod and through the `config.py`. Do other environment variables currently exist in the application? If so, find their example usages, and I can add the variable for you in Travis.

I forgot what you're supposed to do locally, but you could inline the key into `config.py` as long as you don't commit it.

http://flask.pocoo.org/docs/1.0/config/

app/api/routes.py
kylemh: Indexes should start at 0

app/api/routes.py
kylemh: This is the inappropriate place for this. 

Imagine the front-end calls the search end point AFTER a new item is added.

The request will now take a much longer time and will return unexpected results.

The Algolia index should be updated specifically when resources are created, deleted, or updated. This will lead to more code, but it is more testable/expected

Digbigpig: Yeah, I dont want this code to be here, I just kinda left it so that if someone runs this code, it will at least populate their index. 

I attempted to have it only run on startup for now (within app.create_app) but I kept having import errors when I would import Resources. I also tried having it run within cli.register but that was a similar situation.

app/__init__.py
Digbigpig: @kylemh I believe I added the pattern properly. It wont work till travis has the info right? I can give you my info for now if we dont have the OSS account yet

kylemh: I can implement it for you! DM me the key on Slack

app/models.py
aaron-suarez: The quick fix for the tests is to add `'objectID': None,` right here
https://github.com/OperationCode/resources_api/blob/b6104234e3e5e2004b7cb285747815a44aaf29fc/tests/unit/test_models.py#L24-L37

But I don't like that this breaks the convention of everything else in the file. Perhaps we can find a way to inject it after the fact so it doesn't look out of place here?

app/api/routes.py
aaron-suarez: When you do a Google search, it puts your query in a query param called `q`, so the URL ends up something like `www.google.com/search?q=operation+code`. I think it would be good to follow this pattern, especially because I'm not sure how well path parameters will be able to be url encoded (i.e., replacing a space with `+`). Also, the search should be searching all the data (like languages and categories). So I think we should just make it `/search` instead of `/resources/search` and then look in the query params for `q` to find the specific search terms

app/__init__.py
```diff
@@ -11,6 +12,10 @@
 
 API_VERSION = "1.0"
 
+# Connect to Agolia
+client = SearchClient.create('your_project_name', 'your_key')
```
aaron-suarez: Can we call this `search_client` or `agolia_client` perhaps? In the routes.py file it's not entirely clear what service this is a client of. I'm a fan of `search_client` because if we ever change vendors (perhaps one day we move to elastic search?) then we probably won't have to change the name. And if we ever use another client for something completely different, it won't be confused with that, because we know we're talking about the _search_ client specifically.

app/api/routes.py
Digbigpig: I agree with replacing the path parameters with regular params and changing it to /search.

My confusion is, when you say you want them to be able to search languages/categories, are you saying you want it to return a Language/Category object? If that's what you mean, could you explain the use case for that?

Otherwise, language/category is already being used within the search. We can add filters for categories/languages though.

aaron-suarez: I just meant that if someone searches for `python` they're searching for a certain language, so calling it `/resources/search` isn't entirely correct. Although I could see the case for how you only want resources, because like you said, this endpoint will only be returning resources. That being said, it doesn't really matter, you already said you agree with changing it to `/search` which was the point of me saying that, so we're good.

app/__init__.py
```diff
@@ -1,4 +1,5 @@
-from configs import Config
+from algoliasearch.search_client import SearchClient
+from configs import Config, search_user, search_key, search_resource_index
```
aaron-suarez: Rather than importing in `search_user`, `search_key` and `search_resource_index`, why don't we go ahead and add those as constants to the `Config` class?
https://github.com/OperationCode/resources_api/blob/b6104234e3e5e2004b7cb285747815a44aaf29fc/configs.py#L43

Then we could just do `Config.SEARCH_USER` rather than importing them separately. This would be more in keeping with the config patterns of the app.

Digbigpig: Sounds good. 
I considered doing that but was hesitant to mess with that class since it was used with the create_app function.

app/api/routes.py
aaron-suarez: I feel like there has to be a more efficient way to do this. Let's suppose that algolia returns 80 results. This is going to query the database 80 times to get all of the results from the database? That would take way too long

app/api/routes.py
aaron-suarez: You know, you're right. Our pagination does start at 1, I never realized that was weird.

Did a bit of digging, and it turns out that this was a choice by the makers of Flask-SQLAlchemy, which we use as our ORM https://flask-sqlalchemy.palletsprojects.com/en/2.x/api/#flask_sqlalchemy.Pagination.page

I'm not sure why they chose to make it 1 indexed, but translating that to a 0 index would certainly be more work than its worth, even though Kyle is right that it probably ought to be 0 indexed

configs.py
```diff
@@ -32,6 +32,13 @@ def get_sys_exec_root_or_drive():
 if not postgres_host:
     raise KeyError("Application requires 'POSTGRES_HOST' to run")
 
+search_user = os.environ.get('SEARCH_USER')
+search_key = os.environ.get('SEARCH_KEY')
+if not all([search_user, search_key]):
+    print("Application requires 'POSTGRES_HOST' and 'POSTGRES_KEY for search funtionality")
```
aaron-suarez: Should probably print `Application requires 'SEARCH_USER' and 'SEARCH_KEY for search functionality`

requirements.txt
```diff
@@ -1,3 +1,4 @@
+algoliasearch>=2.0,<3.0
```
aaron-suarez: We'll want to pin this to a specific version. I understand that they use semver so we're basically just creating more work for ourselves by pinning it rather than letting it get upgraded to the latest minor version automatically, but I still think that it makes more sense to pin it rather than keep it variable.

@wimo7083 What do you think? Should we specify a more specific version of Algolia or call it good at "between 2 and 3"? FWIW, the Algolia docs specify the way he has it https://www.algolia.com/doc/api-client/getting-started/update-the-client/python/?language=python#with-pip

app/api/routes.py
Digbigpig: You're right, I should write a join and query once.

app/api/routes.py
```diff
@@ -179,6 +184,57 @@ def get_resources():
...
+            break
+
+    search_result = index.search(f'{term}', {
+        'attributesToRetrieve': [
+            'name',
+            'url',
+            'category',
+            'languages',
+            'notes',
+        ],
```
aaron-suarez: If we're just going to get the data from the database here https://github.com/OperationCode/resources_api/blob/b6104234e3e5e2004b7cb285747815a44aaf29fc/app/api/routes.py#L225
then all we need from algolia is the objectid, right? Are we actually doing anything with these attributes?

aaron-suarez:
> I'm assuming we want to add filtering to the search too which I haven't started on yet.
> `Category, language, paid` filters seem like the obvious ones.

Regarding this, I think that we probably want a filter for paid, but I don't think we necessarily need filters for category and language (at least not right away)

app/models.py
Digbigpig: I made another property that calls serialize and adds the objectID to it.

app/utils.py
```diff
@@ -108,3 +108,11 @@ def setup_logger(name, log_file, level=logging.INFO):
...
 
     return logger
+
+
+def reindex_search(query):
+    print("Reindexing Database")
+
+    index.clear_objects()
+    db_list = [u.serialize_algolia_search for u in query.all()]
+    index.save_objects(db_list)
```
Digbigpig: This is the code to rebuild the search index from our database. I haven't implemented it yet because we haven't decided where/how we want this ran. It really shouldn't be ran unless you're starting up and haven't created an index before or the data changed when the server was down.

.env
```diff
@@ -4,3 +4,7 @@ POSTGRES_HOST=resources-postgres
 POSTGRES_DB=resources-psql
 FLASK_ENV=development
 FLASK_DEBUG=1
+SEARCH_USER=search_user
+SEARCH_KEY=search_key
```
aaron-suarez: The [Algolia docs](https://www.algolia.com/doc/api-client/getting-started/instantiate-client-index/#init-index) call this the `Application ID` and `API Key` rather than the Search User and Search Key. Can we rename it such that it matches their terms?

.env
```diff
@@ -4,3 +4,7 @@ POSTGRES_HOST=resources-postgres
...
+SEARCH_RESOURCE_INDEX=search_resource_index
```
aaron-suarez: In the same way, the docs call this an `index name`, can we match that terminology please?

aaron-suarez: Meaning I think it should be `INDEX_NAME=search_resource_index`

app/api/routes.py
```diff
@@ -179,6 +184,32 @@ def get_resources():
...
 
 
+def search_results():
+    term = request.args.get('q', '', str)
+    page = request.args.get('page', 0, int)
+    query = Resource.query
+
+    search_result = index.search(f'{term}', {
+        'page': page,
+    })
```
aaron-suarez: Clients are allowed to pass in a query param called `page_size`, and we should honor that here.

Should add in something like
```python
'hitsPerPage': request.args.get('page_size')
```
but default to the `per_page` specified in the Configs file if they don't pass in this param

app/api/routes.py
```diff
@@ -307,12 +338,18 @@ def create_resource(json, db):
     try:
         db.session.add(new_resource)
         db.session.commit()
+        index.save_object(new_resource.serialize_algolia_search)
```
aaron-suarez: We need to add a similar call on `set_resource` (and to be honest, I don't like the name `set_resource`, we should call it `update_resource` and change the current `update_resource` to `put_resource` so that way it follows the same pattern as `post_resources()` and `create_resource`).

But of course, we should use `index.partial_update_object` and only update those fields that changed (unless its easier to just save the whole object, either way works for me)

.env
```diff
@@ -4,3 +4,7 @@ POSTGRES_HOST=resources-postgres
...
+INDEX_SEARCHED=FALSE
```
aaron-suarez: Should this be `INDEX_CHECKED` instead of `INDEX_SEARCHED`? This doesn't appear to be used anywhere, and `INDEX_CHECKED` doesn't seem to be declared anywhere even though its used

run.py
```diff
@@ -1,9 +1,24 @@
...
 
 app = create_app()
 cli.register(app, db)
 
+with app.app_context():
+    if os.environ.get('INDEX_CHECKED') == "FALSE":
+        query = Resource.query
+        indicies = search_client.list_indices()
+        for ind in indicies['items']:
+            if ind['name'] == 'resources_api':
```
aaron-suarez: Are there two different indexes, one called `resources_api` and another called `search_resource_index`? I find this part rather confusing. Personally, I like the name `resources_api` for the index, and I only think we need one index. We should probably change the `.env` file to reflect this name if that's the one we're going with (and if it's the same index)

run.py
```diff
@@ -1,9 +1,24 @@
...
+                curr_index = ind
+                if curr_index['entries'] != query.all().count():
```
aaron-suarez: Should this actually be `if curr_index['entries'] != query.count():`? I couldn't get it to work without making that change

app/api/routes.py
```diff
@@ -307,12 +338,18 @@ def create_resource(json, db):
...
+
+    except AlgoliaUnreachableHostException as e:
```
aaron-suarez: Annoyingly, `AlgoliaUnreachableHostException` does not extend from `AlgoliaException`, so we have to catch both. Probably can use the same logic you have in there now for the AlgoliaException too

app/api/routes.py
```diff
@@ -179,6 +184,32 @@ def get_resources():
...
+
+    pagination_details = {
+                "pagination_details": {
+                    "page": search_result['page'],
+                    "number_of_pages": search_result['nbPages'],
+                    "records_per_page": search_result['hitsPerPage'],
+                    "total_count": search_result['nbHits'],
+                }
+        }
+    return standardize_response(payload=dict(data=results, **pagination_details))
```
aaron-suarez: We need some slightly better error handling here. If there's an internal server error, you get something similar to the following:
```json
{
    "apiVersion": "1.0",
    "data": null,
    "errors": [
        {
            "code": "something-went-wrong"
        }
    ],
    "number_of_pages": 0,
    "page": 0,
    "records_per_page": 20,
    "status": "error",
    "status_code": 500,
    "total_count": 0
}
```

aaron-suarez: Perhaps the solution to this can go inside `standardize_response` but I'll let you take that in whichever direction you want

app/utils.py
aaron-suarez: Is this supposed to do the same thing as the code you added to `run.py`? If not, what are the differences?

Digbigpig: I've removed this.

.env
Digbigpig: Yeah, that was a typo. Fixed in next commit

run.py
Digbigpig: I made it "search_resource_index" as a placeholder like the rest of the search related variables. 
I will fix this to get the indexes name from the environment variable.

app/api/routes.py
aaron-suarez: To make this more readable, let's make the `request.args.get` call above and store it in a variable like we do with `page`

configs.py
```diff
@@ -32,10 +32,20 @@ def get_sys_exec_root_or_drive():
...
+if not all([search_id, search_key]):
+    print("Application requires 'APPLICATION_ID' and 'API_KEY' for search functionality.")
+
+index_name = os.environ.get("INDEX_NAME")
+
 
 class Config:
     SQLALCHEMY_TRACK_MODIFICATIONS = False
     SQLALCHEMY_DATABASE_URI = f"postgresql://{postgres_user}:{postgres_password}@{postgres_host}:5432/{postgres_db}"
+    SEARCH_ID = search_id
```
aaron-suarez: Let's change these to application id as well, for consistency

configs.py
```diff
@@ -32,10 +32,20 @@ def get_sys_exec_root_or_drive():
...
+    SEARCH_KEY = search_key
```
aaron-suarez: And api key here

run.py
```diff
@@ -1,9 +1,25 @@
...
 
+with app.app_context():
+    if os.environ.get('INDEX_SEARCHED') == "FALSE":
+        query = Resource.query
+        indicies = search_client.list_indices()
+        for ind in indicies['items']:
+            if ind['name'] == os.environ.get('INDEX_NAME'):
+                curr_index = ind
+
+                if curr_index['entries'] != query.count():
```
aaron-suarez: This feels a little brittle still. Do we have to reindex the whole thing if this count is wrong? Also, what if a bunch of resources changed somehow, but it turned out to be the same count? I'm not really sure I know a better way to do it though... @kylemh Do you have any suggestions? Or do you happen to know if this is a best practice?

Digbigpig: One option is so iterate over every resource and see if it exists in the index and all of its properties are correct. If not we can add it to a batch of operations to update the index. It would take longer than the current solution but would cut down on our operations. 

They have a browse method to return our whole index. https://www.algolia.com/doc/api-reference/api-methods/browse/

wimo7083: This needs to not be in the way of the user experience. 

We don't want this to occur in the main application logic. I don't recall if every request has to wait for this but it's in run.py which we forward the requests to. 

How often does this section of code need to be ran?

It looks like this only needs to be done when we are out of sync from algolia and our db. Maybe this only needs to occur on the update resource (and in a background celery task) or even only on every deploy.

aaron-suarez: I don't think it even needs to be run on _every_ deploy. Do you think it needs to be run that often @Digbigpig ?

Digbigpig:
I'm sorry for the mess of commits.

aaron-suarez:
Closing in favor of #172

PR #151, opened 2019-05-06
Pinning pyyaml dependency

soris-codes:
Thanks for the opportunity to learn/read about pinning dependencies! 
Pyyaml's current stable version is 5.1. Not sure if we needed/wanted to move to that one yet.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,6 +4,6 @@ Flask-Migrate==2.4.0
 psycopg2-binary==2.7.7
 pytest-cov==2.6.1
 pytest-mock==1.10.1
-pyyaml==4.2b1
+pyyaml==5.1
 requests==2.21.0
 SQLAlchemy-Utils==0.33.11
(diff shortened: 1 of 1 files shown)


aaron-suarez:
@soris-codes The version of pyyaml we're currently using is 5.1. When it said `pyyaml>=4.2b1`, it defaulted to that version, so it works. As a matter of fact, we're not compatible with 4.2b1 which is why Travis is failing.

requirements.txt
```diff
@@ -4,6 +4,6 @@ Flask-Migrate==2.4.0
 psycopg2-binary==2.7.7
 pytest-cov==2.6.1
 pytest-mock==1.10.1
-pyyaml>=4.2b1
+pyyaml==4.2b1
```
aaron-suarez: Please update this to `pyyaml==5.1` and we'll be good to go!

soris-codes: Updated requirements.txt to show pyyaml==5.1

PR #152, opened 2019-05-07
Cycle Code Climate API key

aaron-suarez:
This removes the API key for code climate from the .travis.yml file and replaces it with the secure, encrypted version of the new API key.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,43 +1,38 @@
 env:
+  matrix:
+  - DOCKER_COMPOSE_VERSION=1.22.0
   global:
-    - SQLALCHEMY_DATABASE_URI="sqlite:///:memory:"
-    - IMAGE_NAME=resources-api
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,43 +1,38 @@
...
-
+- make build
+- make test
+- make lint
 after_success:
-  - bash bin/run.sh
-
+- bash bin/run.sh
 after_script:
-  - docker-compose run resources-api coverage xml
```
wimo7083: Why is this showing a delta?

.travis.yml
```diff
@@ -1,43 +1,38 @@
...
 script:
-  - 'make build'
-  - 'make test'
-  - 'make lint'
-
+- make build
+- make test
+- make lint
 after_success:
-  - bash bin/run.sh
```
wimo7083: Why is this showing a delta?

PR #158, opened 2019-05-20
Add security checks

### LINKED ISSUE #153: Security checks in CI/CD pipeline
We should incorporate Bandit and similar tools into our dev dependencies and CI checks to verify that there are no easily scannable security issues with the project.

aaron-suarez:
Fixes #153 

This adds bandit into the CI pipeline. I went ahead and excluded `/tests` because there were a bunch of issues but I don't think that matters. Because I wanted to exclude `/tests`, I had to specify bandit 1.5.1. We will need to upgrade past 1.6.0 once this is resolved https://github.com/PyCQA/bandit/issues/488 or we won't be able to exclude the tests directory.

### DIFF
--- a/.bandit
+++ b/.bandit
@@ -1,0 +1,2 @@
+[bandit]
+exclude: /tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ script:
 - make build
 - make test
 - make lint
+- make bandit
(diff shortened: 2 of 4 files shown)


lord-cant-even:
seems ok to me, I've never used bandit but it seems to have a nice enough pedigree

PR #174, opened 2019-06-09
Optimize search results

### LINKED ISSUE #173: Do not query database for Algolia results
Regarding https://github.com/OperationCode/resources_api/pull/172#discussion_r291847339 we should not be doing a query on the database, we should trust that the Algolia results are correct and return them to the browser. Any changes in formatting can be made in python rather than by querying the DB. If we find that the algolia results are outdated, we should use the reindex click command to fix the index in prod.

Digbigpig:
This commit has search results displayed directly from algolia
without querying the database first.

Closes #173

### DIFF
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -215,7 +215,6 @@ def search_results():
     term = request.args.get('q', '', str)
     page = request.args.get('page', 0, int)
     page_size = request.args.get('page_size', Config.RESOURCE_PAGINATOR.per_page, int)
-    query = Resource.query
 
     search_result = index.search(f'{term}', {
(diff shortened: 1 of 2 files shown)


app/api/routes.py
```diff
@@ -437,3 +435,22 @@ def validate_resource(json):
 
     if validation_errors['errors']['missing_params']:
         return validation_errors
+
+
+def format_resource_search(hit):
```
aaron-suarez: This file is super long, let's go ahead and stick this in `utils`. We should go ahead and stick `validate_resource` and `create_new_apikey` in resources too while we're at it

PR #182, opened 2019-06-15
Migrate from Travis to CircleCI

AllenAnthes:
Adds circleCI configs and removes `travis.yml`


requirements.txt
```diff
@@ -12,4 +12,4 @@ SQLAlchemy-Utils==0.34.0
 prometheus_client==0.7.0
 uWSGI==2.0.18
 py-healthcheck==1.9.0
-bandit==1.5.1
+bandit==1.5.1
```
aaron-suarez: Nit: can you put this newline back in here please? The linter doesn't catch it for non-python files

PR #185, opened 2019-06-16
Returns 422 when non unique parameters are sent to resource post and put

### LINKED ISSUE #159: Update/Create resource 500's on non unique attribute
The update and create resource endpoint returns a 500 error if there is a non unique constraint in the url field. This should be caught (possibly 'sqlalchemy.exc.IntegrityError') and return a 422 error.

Digbigpig:
Closes #159


Digbigpig:
TODO: Incorporate invalid params for return message.

https://github.com/OperationCode/resources_api/blob/master/app/utils.py#L155

Digbigpig:
TODO: Give invalid params a description so that users know why it is invalid,
            Validate param types. Example: ensure "paid" is a boolean value before attempting to submit to db

app/utils.py
```diff
@@ -152,25 +152,44 @@ def setup_logger(name, level=logging.INFO):
...
+
+    url = json.get("url")
+    if url is not None:
+        # If there is an existing resource with the requested url and it isn't the Resource we are trying to update.
+        resource = Resource.query.filter_by(url=url).first()
+        if resource and resource.id != id:
+            invalid_params.append("url")
+
+    if missing_params:
+        validation_errors['errors']['missing_params'] = []
```
aaron-suarez: I don't think this line is necessary. You're setting `missing_params = []` up on line 158, so the next line will have the same effect if there are no missing params. Same with line 191, seems unnecessary.

app/utils.py
```diff
@@ -152,25 +152,44 @@ def setup_logger(name, level=logging.INFO):
...
+        for prop in required:
+            if json.get(prop) is None:
+                missing_params.append(prop)
+
+    url = json.get("url")
+    if url is not None:
+        # If there is an existing resource with the requested url and it isn't the Resource we are trying to update.
+        resource = Resource.query.filter_by(url=url).first()
+        if resource and resource.id != id:
+            invalid_params.append("url")
```
aaron-suarez: We should have a message in `validation_errors['errors']['message']` that states the url already exists in the database. Otherwise, people will be wondering why it's invalid. And since we're already pulling the resource, we have the id. So we can provide that as well. You'd get a response like this:
```json
{
    "apiVersion": "1.0",
    "data": null,
    "errors": {
        "invalid_params": [
            "url"
        ],
         "message": "This URL already exists in the database. Resource id 135 already has this URL.",
        "type": "validation"
    },
    "status_code": 422
}
```

aaron-suarez:
As we're going through these changes to the error response, let's make sure we're conforming to the standard set described in https://github.com/OperationCode/resources_api/issues/196

In short, each response should have an `errors` attribute which is an array of error objects, each of which contains a `code` and a `message`. The `invalid_params` and `missing_params` need only be present when relevant, but the code and message should always be there so that the front end can rely on those attributes to decide what to do in their error handling.

I'm sure this was already going to be the case, but I just wanted to drop the note as a quick reminder that this is something we need to care about now that it's documented in https://github.com/OperationCode/resource_api_docs/pull/1

app/utils.py
aaron-suarez: In accordance with my latest comment regarding the `code` attribute, I'm updating the above example to include that as well

Digbigpig:
If I remember correctly, the only thing I have left is to add testing around this. I have uncommitted stuff still but I should have some time to finish it up tomorrow.

app/api/routes.py
```diff
@@ -332,7 +338,7 @@ def update_resource(id, json, db):
         return redirect('/404')
 
     langs, categ = get_attributes(json)
-    index_object = {'objectID': resource.id}
+    index_object = {'objectID': id}
```
Digbigpig: @aaron-suarez this and the partial _update_object change lets it update properly. The change to category also lets it save the right category

app/utils.py
```diff
@@ -154,29 +162,141 @@ def setup_logger(name, level=logging.INFO):
...
-    validation_errors = {'errors': {'type': 'validation'}}
+def validate_resource(request, id=-1):
+    json = request.get_json()
+    validation_errors = {'errors': [],
+                         'type': 'validation'}
+
     if not json:
-        message = "A JSON body is required to use this endpoint, but none was given"
-        validation_errors['errors']['message'] = message
+        message = generate_error(ValidationError.MissingJSONError, field=None)
```
aaron-suarez: Nit: is message still a good name for this considering it's now a dict with a bunch of data in it? Maybe we should call this `error`

app/utils.py
```diff
@@ -154,29 +162,141 @@ def setup_logger(name, level=logging.INFO):
...
+            # The provided category will get mapped to this resource later on.
+            if column.name == 'category_id':
+                col_type = str
+
+            # Allow String columns to accept integers.
+            if col_type is str:
+                if type(field) in [int, float]:
+
+                    # Do not allow integers as urls
+                    if column.name.replace('_id', '') != 'url':
```
aaron-suarez: We already declare `name` to be `column.name.replace('_id', '')` up on line 182. We keep calling this replace function over and over again, but can we just use `name` instead?

aaron-suarez: Might be good to rename `name` as `column_name` or `param_name` or something because maybe `name` is too generic. But that's not a huge deal, we can leave it as `name` for now

app/utils.py
```diff
@@ -154,29 +162,141 @@ def setup_logger(name, level=logging.INFO):
...
+                    generate_error(ValidationError.TypeError,
+                                   column.name.replace('_id', ''),
+                                   received=type(field).__name__,
+                                   expected=col_type.__name__)
+                )
+
+    # Special case for Languages - Must be list and all elements must be str.
+    langs = json.get('languages')
+    if langs:
+        if type(langs) is not list or not all(map(lambda x: type(x) is str, langs)):
```
aaron-suarez: I love python ❤️ reads like english

app/utils.py
```diff
@@ -154,29 +162,141 @@ def setup_logger(name, level=logging.INFO):
...
-    if validation_errors['errors']['missing_params']:
+            validation_errors['errors'].append(
+                generate_error(ValidationError.MissingParamError, prop))
+
+    url = json.get("url")
+    if url is not None:
+
+        # If there is an existing resource with the requested url
+        # and it isn't the Resource we are trying to update, return an error.
+        resource = Resource.query.filter_by(url=url).first()
```
aaron-suarez: Isn't this unnecessary now that we're catching the `IntegrityError` when we try to actually create/update the record in the database? It would be better to just have the one query to insert it and return the `ValidationError.IntegrityError` from the catch block in the routes file, right?

app/utils.py
```diff
@@ -154,29 +162,141 @@ def setup_logger(name, level=logging.INFO):
     return logger
 
 
-def validate_resource(json):
-    validation_errors = {'errors': {'type': 'validation'}}
+def validate_resource(request, id=-1):
```
aaron-suarez: This function has gotten pretty long and complex. I think we should break it down into parts and call all of those as separate functions. Like one function to check that there's no missing params, another function to validate the type, etc. Just something to break it down a bit.

app/utils.py
Digbigpig: Good catch.

aaron-suarez: Upon further discussion in slack, @Digbigpig and I decided that getting rid of it is premature optimization. We should still catch the `IntegrityError` and return a 422 in that case, but we should stop it before it gets that far. If later we decide that we want to implement caching to prevent making 2 db calls in subsequent requests, we'll consider that when we find that it's necessary. "Premature optimization is the root of all evil"

app/api/validations.py
```diff
@@ -0,0 +1,109 @@
...
+    return wrapper
+
+
+def missing_json_error():
+    message = "You must provide a valid JSON body to use this endpoint"
+    error = {'errors': {"missing-body":{"message":message}}}
+    return standardize_response(error, status_code=422)
+
+
+def validate_resource(request, id=-1):
```
aaron-suarez: TODO: I think we could probably break this down into some smaller functions

app/api/routes.py
```diff
@@ -389,6 +405,10 @@ def update_resource(id, json, db):
             payload=dict(data=resource.serialize)
             )
 
+    except IntegrityError as e:
```
aaron-suarez: TODO: we need to get some coverage around this

app/utils.py
```diff
@@ -106,29 +106,41 @@ def standardize_response(payload={}, status_code=200):
...
-                "code": '-'.join(err_map.get(status_code).split(' ')).lower()
-            }]
+            code = '-'.join(err_map.get(status_code).split(' ')).lower()
+            message = msg_map.get(status_code)
+            resp["errors"] = {}
+            resp["errors"][code] = {"message": message}
 
     elif not data:
-        resp["errors"] = [{"code": "something-went-wrong"}]
+        resp["errors"] = {'errors': {"server-error":{"message":"Something went wrong"}}}
```
aaron-suarez: TODO: we need coverage around this code

aaron-suarez:
TODO: Add all these changes to openapi.yaml so they make it into the docs

app/api/validations.py
```diff
@@ -0,0 +1,109 @@
...
+                return missing_json_error()
+
+        return func(*args, **kwargs)
+    return wrapper
+
+
+def missing_json_error():
+    message = "You must provide a valid JSON body to use this endpoint"
+    error = {'errors': {"missing-body":{"message":message}}}
+    return standardize_response(error, status_code=422)
```
aaron-suarez: TODO: Add all these changes (like all of these breaking changes) to the openapi.yaml file so they make it into the docs

aaron-suarez:
Ok @Digbigpig it's getting late, I can't keep working on this. If you would, please go ahead and pull down my changes (I recommend deleting your local branch, fetching this one, and then switching over to it -- do not attempt to merge this commit into your branch!). Once you have the new branch, see if you can break anything or if any of the responses don't look right per our conversation earlier on slack.

If everything looks good, I could really use some help on those `TODO` items I posted above, particularly on adding everything to the openapi.yaml file. I haven't gotten a chance to look at the version you sent me, but I imagine it's out of date because of all the changes I made to the responses today. Please go ahead and update those, then I can review that and we can get this merged!

If you find any problems while you're playing with it, or feel like something could be better in the code, feel free to either go ahead and fix it or to leave a review comment and I can fix it. I'm not sure how much time over the next two weeks I'll have, but I'll do my best to tag team you on this. I know the frontend people are ready to start building out the resources page, and I'd like to unblock them by not having any more breaking changes (or at least versioning our breaking changes) once they get started.

Thanks for your help on this!!

app/api/validations.py
Digbigpig: IMO, there's not a great way to break this down, that will add value to the code.

app/utils.py
Digbigpig: This one is kind of weird... I think I'll have a test that mocks the standarized response itselt and just returns standardized_response(1337) which will fail the if statement.

Digbigpig:
image
I added the tests we wanted to make. I'm not fully satisfied with them but I wanted to commit what I got done.

app/static/openapi.yaml
```diff
@@ -69,10 +69,13 @@ paths:
                 {
                   'apiVersion': '1.0',
                   'data': null,
-                  'errors': [{
-                      'code': 'unprocessable-entity',
-                      'message': 'Your username or password is incorrect',
-                  }],
+                  'errors': {
+                    'invalid-params': {
```
aaron-suarez: Can we also add the `missing-params` example in here?

app/static/openapi.yaml
```diff
@@ -943,11 +947,10 @@ components:
...
               'apiVersion': '1.0',
               'data': null,
-              'errors': [
-                  {
-                      'code': 'not-found'
-                  }
-              ],
+              'errors': {
+                'not-found': {
+                  'message': "The resource you requested does not exist.'
```
aaron-suarez: This should be `'The resource...` instead of `"The resource...` because it renders with a `\"` when you use the double quotes

app/static/openapi.yaml
```diff
@@ -620,8 +624,8 @@ components:
           type: object
           description: A data object
         errors:
-          type: array
-          description: List of errors
+          type: object
+          description: The response errors
```
aaron-suarez: We should show what the error object looks like. We do this for the `data` object in a successful response here https://github.com/OperationCode/resources_api/blob/52a1c4b805b8913baa4af58ac67ddf22ba37afd3/app/static/openapi.yaml#L783-L788

Perhaps create a `Error_data` and list out the properties there? We want to make sure all of the properties are available to view.

app/static/openapi.yaml
```diff
@@ -962,11 +965,11 @@ components:
...
               'apiVersion': '1.0',
               'data': null,
-              'errors': [
-                  {
-                      'code': 'unauthorized'
-                  }
-              ],
+              'errors': {
+                'invalid_credentials': {
+                  'message': 'The email or password you submitted is incorrect'
```
aaron-suarez: The 401 response sometimes looks like this:
```json
{
    "apiVersion": "1.0",
    "data": null,
    "errors": {
        "unauthorized": {
            "message": "You must provide a valid API token in the x-apikey header."
        }
    },
    "status": "Unauthorized",
    "status_code": 401
}
```
I don't think that we can reuse this same 401 response example for all of the endpoints anymore. We need to have this `invalid_credentials` example for the Request Key endpoint and the one I pasted above for the Create/Update Resource endpoints

aaron-suarez:
You didn't make a change to this part of the openapi.yaml file, but we need to add a 422 response to the `List Resources` endpoint because if you have an `updated_after` in the future it will return a 422.

PR #186, opened 2019-06-16
Added testing for the search endpoint

### LINKED ISSUE #183: Need tests around search endpoint
Assigning @Digbigpig because he was the person who opened the original PR for the search endpoint. If I get to this before him, that's fine.

There is no automated test in the /test directory around the search endpoint. We'll need to mock the algolia service and test the route.

Digbigpig:
Closes #183 
Screenshot from 2019-06-16 13-11-07

### DIFF
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -368,7 +368,6 @@ def update_resource(id, json, db):
         except (AlgoliaUnreachableHostException, AlgoliaException) as e:
             logger.exception(e)
             print(f"Algolia failed to update index for resource '{resource.name}'")
-            print(f'Update data: {index_object}')
 
         return utils.standardize_response(
             payload=dict(data=resource.serialize)
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -218,10 +218,15 @@ def search_results():
     page = request.args.get('page', 0, int)
     page_size = request.args.get('page_size', Config.RESOURCE_PAGINATOR.per_page, int)
 
-    search_result = index.search(f'{term}', {
-        'page': page,
-        'hitsPerPage': page_size
-    })
+    try:
+        search_result = index.search(f'{term}', {
+            'page': page,
+            'hitsPerPage': page_size
+        })
+
+    except (AlgoliaUnreachableHostException, AlgoliaException) as e:
+        logger.exception(e)
+        return utils.standardize_response(status_code=500)
 
     if page >= int(search_result['nbPages']):
         return redirect('/404')
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -332,7 +332,7 @@ def update_resource(id, json, db):
         return redirect('/404')
 
     langs, categ = get_attributes(json)
-    index_object = {'objectID': resource.id}
+    index_object = {'objectID': id}
 
     try:
         logger.info(f"Updating resource. Old data: {resource.serialize}")
@@ -341,7 +341,7 @@ def update_resource(id, json, db):
             index_object['languages'] = resource.serialize['languages']
         if json.get('category'):
             resource.category = categ
-            index_object['categories'] = categ
+            index_object['category'] = categ.name
         if json.get('name'):
             resource.name = json.get('name')
             index_object['name'] = json.get('name')
@@ -358,11 +358,12 @@ def update_resource(id, json, db):
         db.session.commit()
 
         try:
-            index.partial_update_objects(index_object)
+            index.partial_update_object(index_object)
 
         except (AlgoliaUnreachableHostException, AlgoliaException) as e:
             logger.exception(e)
             print(f"Algolia failed to update index for resource '{resource.name}'")
+            print(f'Update data: {index_object}')
 
         return utils.standardize_response(
             payload=dict(data=resource.serialize)
--- a/app/cli.py
+++ b/app/cli.py
@@ -124,7 +124,7 @@ def update_resource(resource, existing_resource):   # pragma: no cover
     existing_resource.languages = resource['languages']
 
 
-def reindex_all():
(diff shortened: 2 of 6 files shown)


tests/unit/test_routes.py
```diff
@@ -371,6 +371,66 @@ def test_update_resource(module_client, module_db, fake_auth_from_oc):
     assert (response.status_code == 404)
 
 
+def test_search(module_client, module_db, fake_auth_from_oc):
```
aaron-suarez: `/search` isn't an authenticated endpoint, do we need the mock for fake_auth_from_oc?

requirements.txt
```diff
@@ -2,6 +2,7 @@ algoliasearch>=2.0,<3.0
 flake8==3.7.7
 Flask-Limiter==1.0.1
 Flask-Migrate==2.5.2
+names==0.3.0
```
aaron-suarez: Although this looks like a cool little project, I don't think it's necessary for us to pull in another dependency. Can we just generate random strings using something like this?

```python
import string
import random

name = ''.join(random.choices(string.ascii_uppercase + string.digits, k=10))
```
If you want it to be longer or shorter than 10 characters, you can change that last number to whatever length you want. If we're not testing based on numbers, this would be even shorter:
```python
''.join(random.choices(string.ascii_uppercase, k=10))
```

tests/unit/test_routes.py
aaron-suarez: Unit tests should not test outside services. Let's mock the Algolia API calls like we do with the auth from oc here

aaron-suarez: Oh, and I'm glad we discovered this. We're not mocking the indexing of creating and updating the resources in the other tests, are we? We need to mock algolia in any test that performs an algolia operation.

tests/unit/test_routes.py
```diff
@@ -371,6 +371,66 @@ def test_update_resource(module_client, module_db, fake_auth_from_oc):
...
+        time.sleep(.5)
+        result = client.get(f"/api/v1/search?q={first_term}")
+
+    assert (resource.status_code == 200)
+    assert (result.status_code == 200)
+
+    found = False
+    for entry in result.json['data']:
+        if entry.get('url') == resource.json['data'].get('url'):
+            found = True
```
aaron-suarez: I feel like there's a more pythonic way to do these checks. Once we get the mocking down, I'm going to play with this and see if I can come up with a bit of a better way to see if the resource's URL is in the search results. Maybe this is the best way to do it, but maybe not

Digbigpig: After mocking the search, this is pretty irrelevant since we will only return the results we want. We could just do `assert (results['data'][0]['url'] == resource.json['data'].get('url')` for O(1) time if you were concerned with iterating.

Digbigpig:
To-do: Add expected exception testing around the search endpoint.

app/api/routes.py
```diff
@@ -218,10 +218,15 @@ def search_results():
     page = request.args.get('page', 0, int)
     page_size = request.args.get('page_size', Config.RESOURCE_PAGINATOR.per_page, int)
 
-    search_result = index.search(f'{term}', {
-        'page': page,
-        'hitsPerPage': page_size
-    })
+    try:
```
Digbigpig: I wasn't catching the actual search function if it failed so I'm making it return a server error if it does.

app/cli.py
```diff
@@ -124,7 +124,7 @@ def update_resource(resource, existing_resource):   # pragma: no cover
     existing_resource.languages = resource['languages']
 
 
-def reindex_all():
+def reindex_all():  # pragma: no cover
```
Digbigpig: I removed this from coverage since it is a cli command.

aaron-suarez: Is it possible to just ignore this whole file from coverage if we're not worried about coverage for the cli?

tests/unit/test_routes.py
```diff
@@ -371,7 +373,122 @@ def test_update_resource(module_client, module_db, fake_auth_from_oc):
...
+                          headers={'x-apikey': apikey}
+                          )
+    result = client.get(f"/api/v1/search?q={first_term}")
+
+    assert (resource.status_code == 200)
+    assert (result.status_code == 200)
+    assert (result.json['data'][0]['url'] == resource.json['data'].get('url'))
+
+    # Update the resource and test that search results reflect changes
+    updated_term = ''.join(random.choices(string.ascii_uppercase + string.digits, k=10))
```
aaron-suarez: Since we're using this more than once, why not make a utility function called `random_string()`? That would be a bit more readable because you don't have to try to grok how it's creating the random string to know that that's what it's doing.

tests/unit/test_routes.py
```diff
@@ -371,7 +373,122 @@ def test_update_resource(module_client, module_db, fake_auth_from_oc):
...
+def test_algolia_exception_error(module_client, module_db, fake_auth_from_oc, fake_algolia_exception):
+    client = module_client
+    first_term = ''.join(random.choices(string.ascii_uppercase + string.digits, k=10))
+    apikey = get_api_key(client)
+
+    result = client.get(f"/api/v1/search?q=python")
+
+    assert (result.status_code == 500)
+
+    resource = client.post("/api/v1/resources",
```
aaron-suarez: I'm not sure I understand what the purpose is behind this API call and the next one for `put`. We already have a `test_create_resource` and a `test_update_resource` that create and update a resource and check that the status code is 200.

Is the purpose to demonstrate that even when algolia throws an exception, we still return a 200? I think we should also test that something else happens to notify us that this is going on, since we're going to want to fix it but we don't want the Resources API to go down just because algolia is down

PR #189, opened 2019-06-19
Use docker orb

AllenAnthes:
Uses [Docker CircleCI Orb](https://circleci.com/orbs/registry/orb/circleci/docker) for publishing new docker images instead of a custom script.


aaron-suarez:
Let's go ahead and delete the `/bin` folder if it's no longer necessary.

Issue #197, opened 2019-07-23
Need GET /languages/{id} Route


soris-codes:
Currently getting a 404 for the `/languages/{id} endpoint.

aaron-suarez:
Closed in #200

PR #200, opened 2019-07-27
Adding get_language route and tests

soris-codes:
Followed same pattern as get_category route to set up the get_language route and added tests for getting a single language by ID and testing for out of bounds languages.
Tested retrieving a single language on Postman successfully.


soris-codes:
Fixes #197

Issue #198, opened 2019-07-23
Need GET /categories/{id} Route


soris-codes:
Currently getting a 404 for the `/categories/{id} endpoint.

seoulstice:
I'd like to tackle this issue if no one has claimed it yet.

seoulstice:
@aaron-suarez @soris-codes

aaron-suarez:
Yeah go ahead and get it. I'll be happy to review the PR when you get it in

seoulstice:
Thanks!

aaron-suarez:
Closed with #208

PR #208, opened 2019-07-31
Created GET /categories/{id} Route 

seoulstice:
### Description
* Created happy and sad path tests for GET /categories/{id} endpoint
* Created API endpoint for GET /categories/{id}
* Created get_category(id) method

### Type of change
* New feature (non-breaking change which adds functionality)

Closes Issue #198

### DIFF
--- a/tests/unit/test_routes.py
+++ b/tests/unit/test_routes.py
@@ -256,21 +256,25 @@ def test_get_single_category(module_client, module_db):
 
     category = response.json['data']
     assert (isinstance(category.get('name'), str))
-    assert (category.get('name') == 'Tutorials')
     assert (category.get('id') == 4)
 
 
 def test_get_single_category_out_of_bounds(module_client, module_db):
     client = module_client
 
+    # Check GET request to invalid category id - too low
     too_low = 0
-    too_high = 9999
     response = client.get(f"api/v1/categories/{too_low}", follow_redirects=True)
-
     assert (response.status_code == 404)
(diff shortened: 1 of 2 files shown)


tests/unit/test_routes.py
```diff
@@ -246,6 +246,33 @@ def test_categories(module_client, module_db):
...
+    client = module_client
+
+    response = client.get('api/v1/categories/4')
+
+    # Status should be OK
+    assert (response.status_code == 200)
+
+    category = response.json['data']
+    assert (isinstance(category.get('name'), str))
+    assert (category.get('name') == 'Tutorials')
```
aaron-suarez: We don't need this assertion, it too tightly couples the data with the test. If we were to change the yaml file, for example, `Tutorials` might not end up as the fourth category. Just checking that it's a string with ID 4 is fine

aaron-suarez:
Can we please get one more check in addition to the `too_low` and `too_high` that verifies a 404 when you send a string of non-numeric characters? I think we used `waffles` in another test, but it would be nice to take this opportunity to get some strings with other ascii characters and see how it behaves in that situation.

aaron-suarez:
Other than that, I'm good with the change! Thanks so much for the contribution! Let's make it happen

seoulstice:
Made the suggested changes and thanks for the welcome to OC.  This is my first time contributing to open source and my first time working with Python/Docker 👍

PR #215, opened 2019-09-12
Add support for creating multiple resources to POST /resources endpoint

### LINKED ISSUE #206: Batch create of resources
We should allow batch creation of resources by accepting an array in the JSON for the Create Resource endpoint. I have over 100 resources that I need to add and I don't really fancy adding them one by one. The endpoint should return the list of created resources. I haven't really thought through validation all the way (does it need to create each one by one and then tell you which resources are invalid, or does it fail to create any until you submit a valid list?)

aaron-suarez:
Fixes #206 

**EDIT:** The below examples in this description will be updated as I make changes to the interface, so assume that it is accurate and correct based on the latest commit. 

**EDIT 2:** There is an additional breaking change where when Algolia fails, it will now say in the response that the error has to do with Algolia. Now if Algolia fails, the request fails altogether so we always stay in sync. In addition, `PUT /resource/<id>` and `POST /resources` work without needing algolia as long as you're in the development environment.

This PR allows you to send a list of resources to the `POST /resources` endpoint to add all of them at once. If any resource in the list fails validation, none should be created until the issues are fixed, allowing for subsequent similar requests but with the issues fixed.

The happy path looks like this.

Request:
```json
[
    {
    	"name": "Legit new resource",
    	"url": "legitresource.com",
    	"category": "Computer Science",
    	"paid": false,
        "notes": ""
    },
    {
    	"name": "Another Legit new resource",
    	"url": "anotherlegitresource.com",
    	"category": "Computer Science",
    	"paid": false,
        "notes": ""
    }
]
```
Response:
```json
{
  "apiVersion": "1.0",
  "data": [
    {
      "category": "Computer Science",
      "created_at": "2019-09-12 04:10:07",
      "downvotes": 0,
      "id": 2139,
      "languages": [],
      "last_updated": "",
      "name": "Legit new resource",
      "notes": "",
      "paid": false,
      "times_clicked": 0,
      "upvotes": 0,
      "url": "legitresource.com"
    },
    {
      "category": "Computer Science",
      "created_at": "2019-09-12 04:10:07",
      "downvotes": 0,
      "id": 2140,
      "languages": [],
      "last_updated": "",
      "name": "Another Legit new resource",
      "notes": "",
      "paid": false,
      "times_clicked": 0,
      "upvotes": 0,
      "url": "anotherlegitresource.com"
    }
  ],
  "status": "ok",
  "status_code": 200
}
```

The sad path of both failing validation looks like this.

Request:
```json
[
    {
    	"name": "Legit new resource",
        "notes": ""
    },
    {
    	"name": "Another Legit new resource",
    	"url": "anotherlegitresource.com"
    }
]
```
Response:
```json
{
  "apiVersion": "1.0",
  "data": null,
  "errors": [
    {
      "index": 0,
      "missing-params": {
        "message": " The following params were missing: url, category, paid.",
        "params": [
          "url",
          "category",
          "paid"
        ]
      }
    },
    {
      "index": 1,
      "invalid-params": {
        "message": "The following params were invalid: url. Resource id 2140 already has this URL.",
        "params": [
          "url"
        ],
        "resource": "https://resources.operationcode.org/api/v1/2140"
      },
      "missing-params": {
        "message": " The following params were missing: category, paid.",
        "params": [
          "category",
          "paid"
        ]
      }
    }
  ],
  "status_code": 422
}
```
Notice how the list of errors gives an index number. This index maps to the index number of the list that was submitted. The first one in this example request was missing `url`, `category` and `paid`, so it describes that. The second one also had a URL, so that failed validation for even more reasons.

The sad path of only the second one failing validation looks like this.

```json
{
  "apiVersion": "1.0",
  "data": null,
  "errors": [
    {
      "index": 1,
      "invalid-params": {
        "message": "The following params were invalid: url. Resource id 2140 already has this URL.",
        "params": [
          "url"
        ],
        "resource": "https://resources.operationcode.org/api/v1/2140"
      },
      "missing-params": {
        "message": " The following params were missing: category, paid.",
        "params": [
          "category",
          "paid"
        ]
      }
    }
  ],
  "status_code": 422
}
```

As you can see, if the first resource is fine, then `index` is not a key with an error object in the response. The only resource of the batch that failed validation is index 1 (i.e., the second one in the list).

The API now responds with a specific error when you submit the wrong type to a create or update endpoint. It can recognize when the user submits an "object", "int", "array", "number", "boolean", or "string". If any of these are submitted other than the correct type expected, it will respond with the expectation, as shown below. 

The sad path of submitting an array to the update endpoint:
```json
{
  "apiVersion": "1.0",
  "data": null,
  "errors": {
    "invalid-type": {
      "message": "Expected resource object, but found array"
    }
  },
  "status_code": 422
}
```

The sad path of submitting an object to the create endpoint:
```json
{
  "apiVersion": "1.0",
  "data": null,
  "errors": {
    "invalid-type": {
      "message": "Expected list of resources objects, but found object"
    }
  },
  "status_code": 422
}
```

The sad path of submitting too many resources (more than 200) looks like this:
```json
{
  "apiVersion": "1.0",
  "data": null,
  "errors": [
    {
      "too-long": {
        "message": "This endpoint will accept a max of 200 resources"
      }
    }
  ],
  "status_code": 422
}
```

**Algolia errors**

`GET /search` Response if Algolia fails:
```json
{
  "apiVersion": "1.0",
  "data": null,
  "errors": [
    {
      "algolia-failed": {
        "message": "Failed to get resources from Algolia"
      }
    }
  ],
  "status": "Server Error",
  "status_code": 500
}
```

`PUT /resources/<id>` Response if Algolia fails:
```json
{
  "apiVersion": "1.0",
  "data": null,
  "errors": [
    {
      "algolia-failed": {
        "message": "Algolia failed to update index for resource 'Name of Resource'"
      }
    }
  ],
  "status": "Server Error",
  "status_code": 500
}
```

`POST /resources` Response if Algolia fails:
```json
{
  "apiVersion": "1.0",
  "data": null,
  "errors": [
    {
      "algolia-failed": {
        "message": "Algolia failed to index resources"
      }
    }
  ],
  "status": "Server Error",
  "status_code": 500
}
```

### DIFF
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -4,13 +4,15 @@ from sqlalchemy.exc import IntegrityError
 from algoliasearch.exceptions import AlgoliaUnreachableHostException, AlgoliaException
 from app.api import bp
 from app.api.auth import is_user_oc_member, authenticate
-from app.api.validations import validate_resource, requires_body
+from app.api.validations import validate_resource, validate_resource_list, \
+    requires_body, wrong_type
 from app.models import Language, Resource, Category, Key
 from app import Config, db, index
 from dateutil import parser
 from datetime import datetime
 from prometheus_client import Counter, Summary
 import app.utils as utils
+from os import environ
 
 
 # Metrics
@@ -30,15 +32,21 @@ def resources():
 
 @latency_summary.time()
 @failures_counter.count_exceptions()
-@bp.route('/resources', methods=['POST'], endpoint='create_resource')
+@bp.route('/resources', methods=['POST'], endpoint='create_resources')
 @requires_body
 @authenticate
 def post_resources():
-    validation_errors = validate_resource(request)
+    json = request.get_json()
+
+    if not isinstance(json, list):
+        return wrong_type("list of resources objects", type(json))
+
+    validation_errors = validate_resource_list(request.method, json)
 
     if validation_errors:
         return utils.standardize_response(payload=validation_errors, status_code=422)
-    return create_resource(request.get_json(), db)
+
+    return create_resources(json, db)
 
 
 @latency_summary.time()
@@ -54,10 +62,16 @@ def resource(id):
 @requires_body
 @authenticate
 def put_resource(id):
-    validation_errors = validate_resource(request, id)
+    json = request.get_json()
+
+    if not isinstance(json, dict):
+        return wrong_type("resource object", type(json))
+
+    validation_errors = validate_resource(request.method, json, id)
 
     if validation_errors:
-        return utils.standardize_response(payload=validation_errors, status_code=422)
+        errors = {"errors": validation_errors}
+        return utils.standardize_response(payload=errors, status_code=422)
     return update_resource(id, request.get_json(), db)
 
 
@@ -465,6 +479,30 @@ def update_resource(id, json, db):
         return utils.standardize_response(status_code=500)
 
 
+def create_resources(json, db):
+    try:
+        created_resources = []
+
+        for resource in json:
+            res = create_resource(resource, db)
+
+            # Pass along standardized responses generated from create_resource
+            # such as the result of `utils.standardize_response(status_code=500)`
+            if isinstance(res, tuple):
+                return res
+
+            # If it's anything other than a resource, something went wrong so bail here
+            if not isinstance(res, dict):
+                raise Exception(res)
+
+            created_resources.append(res)
+
+        return utils.standardize_response(payload=dict(data=created_resources))
+    except Exception as e:
+        logger.exception(e)
+        return utils.standardize_response(status_code=500)
+
+
 def create_resource(json, db):
     langs, categ = get_attributes(json)
     new_resource = Resource(
@@ -480,16 +518,21 @@ def create_resource(json, db):
         db.session.commit()
         index.save_object(new_resource.serialize_algolia_search)
 
-    except (AlgoliaUnreachableHostException, AlgoliaException) as e:
-        logger.exception(e)
-        print(f"Algolia failed to index new resource '{new_resource.name}'")
-
     except IntegrityError as e:
         logger.exception(e)
         return utils.standardize_response(status_code=422)
+
+    except (AlgoliaUnreachableHostException, AlgoliaException) as e:
+        if (environ.get("FLASK_ENV") != 'development'):
+            db.session.rollback()
+            logger.exception(e)
+            msg = f"Algolia failed to index new resource '{new_resource.name}'"
+            print(msg)
+            error = {'errors': [{"algolia-failed": {"message": msg}}]}
+            return utils.standardize_response(payload=error, status_code=500)
 
     except Exception as e:
         logger.exception(e)
         return utils.standardize_response(status_code=500)
 
-    return utils.standardize_response(payload=dict(data=new_resource.serialize))
+    return new_resource.serialize
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -300,7 +300,10 @@ def search_results():
 
     except (AlgoliaUnreachableHostException, AlgoliaException) as e:
         logger.exception(e)
-        return utils.standardize_response(status_code=500)
+        msg = "Failed to get resources from Algolia"
+        logger.warn(msg)
+        error = {'errors': [{"algolia-failed": {"message": msg}}]}
+        return utils.standardize_response(payload=error, status_code=500)
 
     if page >= int(search_result['nbPages']):
         return redirect('/404')
@@ -457,18 +460,22 @@ def update_resource(id, json, db):
             resource.notes = json.get('notes')
             index_object['notes'] = json.get('notes')
 
-        db.session.commit()
-
         try:
             index.partial_update_object(index_object)
 
         except (AlgoliaUnreachableHostException, AlgoliaException) as e:
-            logger.exception(e)
-            print(f"Algolia failed to update index for resource '{resource.name}'")
+            if (environ.get("FLASK_ENV") != 'development'):
+                logger.exception(e)
+                msg = f"Algolia failed to update index for resource '{resource.name}'"
+                logger.warn(msg)
+                error = {'errors': [{"algolia-failed": {"message": msg}}]}
+                return utils.standardize_response(payload=error, status_code=500)
+
+        # Wait to commit the changes until we know that Aloglia was updated
+        db.session.commit()
 
         return utils.standardize_response(
-            payload=dict(data=resource.serialize)
-            )
+            payload=dict(data=resource.serialize))
 
     except IntegrityError as e:
         logger.exception(e)
@@ -481,58 +488,61 @@ def update_resource(id, json, db):
 
 def create_resources(json, db):
     try:
+        # Resources to return in the response
         created_resources = []
+        # Serialized Resources to send to Algolia
+        created_resources_algolia = []
+        # Resource IDs to delete if Algolia fails to index
+        resource_id_cache = []
 
+        # Create each Resource in the database one by one
         for resource in json:
-            res = create_resource(resource, db)
+            langs, categ = get_attributes(resource)
+            new_resource = Resource(
+                name=resource.get('name'),
+                url=resource.get('url'),
+                category=categ,
+                languages=langs,
+                paid=resource.get('paid'),
+                notes=resource.get('notes'))
 
-            # Pass along standardized responses generated from create_resource
-            # such as the result of `utils.standardize_response(status_code=500)`
-            if isinstance(res, tuple):
-                return res
+            try:
+                db.session.add(new_resource)
+                db.session.commit()
+                created_resources_algolia.append(new_resource.serialize_algolia_search)
+                resource_id_cache.append(new_resource.id)
 
-            # If it's anything other than a resource, something went wrong so bail here
-            if not isinstance(res, dict):
-                raise Exception(res)
+            except IntegrityError as e:
+                logger.exception(e)
+                return utils.standardize_response(status_code=422)
 
-            created_resources.append(res)
+            except Exception as e:
+                logger.exception(e)
+                return utils.standardize_response(status_code=500)
 
+            created_resources.append(new_resource.serialize)
+
+        # Take all the created resources and save them in Algolia with one API call
+        try:
+            index.save_objects(created_resources_algolia)
+
+        except (AlgoliaUnreachableHostException, AlgoliaException) as e:
+            if (environ.get("FLASK_ENV") != 'development'):
+                logger.exception(e)
+
+                # Remove created resources from the db to stay in sync with Algolia
+                for res_id in resource_id_cache:
+                    res = Resource.query.get(res_id)
+                    res.languages.clear()
+                    db.session.delete(res)
+                db.session.commit()
+
+                msg = "Algolia failed to index resources"
+                error = {'errors': [{"algolia-failed": {"message": msg}}]}
+                return utils.standardize_response(payload=error, status_code=500)
+
+        # Success
         return utils.standardize_response(payload=dict(data=created_resources))
     except Exception as e:
         logger.exception(e)
         return utils.standardize_response(status_code=500)
-
-
-def create_resource(json, db):
-    langs, categ = get_attributes(json)
-    new_resource = Resource(
-        name=json.get('name'),
-        url=json.get('url'),
-        category=categ,
-        languages=langs,
-        paid=json.get('paid'),
-        notes=json.get('notes'))
-
-    try:
-        db.session.add(new_resource)
-        db.session.commit()
-        index.save_object(new_resource.serialize_algolia_search)
-
-    except IntegrityError as e:
-        logger.exception(e)
-        return utils.standardize_response(status_code=422)
-
-    except (AlgoliaUnreachableHostException, AlgoliaException) as e:
-        if (environ.get("FLASK_ENV") != 'development'):
-            db.session.rollback()
-            logger.exception(e)
-            msg = f"Algolia failed to index new resource '{new_resource.name}'"
-            print(msg)
-            error = {'errors': [{"algolia-failed": {"message": msg}}]}
-            return utils.standardize_response(payload=error, status_code=500)
-
-    except Exception as e:
-        logger.exception(e)
-        return utils.standardize_response(status_code=500)
-
-    return new_resource.serialize
--- a/app/api/validations.py
+++ b/app/api/validations.py
@@ -10,7 +10,7 @@ INVALID_PARAMS = "invalid-params"
 def requires_body(func):
     def wrapper(*args, **kwargs):
         try:
-            # JSON body is {}
+            # JSON body is {} or []
             if not request.get_json():
                 return missing_json_error()
         except Exception as e:
@@ -24,14 +24,31 @@ def requires_body(func):
 
 def missing_json_error():
     message = "You must provide a valid JSON body to use this endpoint"
-    error = {'errors': {MISSING_BODY: {"message": message}}}
+    error = {'errors': [{MISSING_BODY: {"message": message}}]}
     return standardize_response(error, status_code=422)
 
 
-def validate_resource(request, id=-1):
+def validate_resource_list(method, rlist):
+    errors = {'errors': []}
+    max_resources = 200
+
+    if len(rlist) > max_resources:
+        msg = f"This endpoint will accept a max of {max_resources} resources"
+        return {"errors": [{"too-long": {"message": msg}}]}
+
+    for i, r in enumerate(rlist):
+        validation = validate_resource(method, r)
+        if validation:
+            validation['index'] = i
+            errors['errors'].append(validation)
+
+    if bool(errors['errors']):
+        return errors
+
+
+def validate_resource(method, json, id=-1):
     errors = None
-    json = request.get_json()
-    validation_errors = {"errors": {}}
+    validation_errors = {}
     missing_params = {"params": []}
     invalid_params = {"params": []}
     required = []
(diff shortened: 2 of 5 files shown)


aaron-suarez:
This is still WIP pending adding unit tests. I wanted to get the API changes in front of y'all before I implement better error handling because if we disagree about how this should work I want to hash that out now before I write a bunch of unit tests to validate this behavior which may change drastically after the conversation.

So please take a look, give any comments about the API or the general code structure, and once we come to an agreement about what this should look like (in both request/response as well as code) I'll fix the issues and write unit tests. Thanks in advance!

app/api/routes.py
```diff
@@ -430,6 +440,27 @@ def update_resource(id, json, db):
         return utils.standardize_response(status_code=500)
 
 
+def create_resources(json, db):
+    try:
+        if isinstance(json, dict):
+            return utils.standardize_response(payload=dict(
+                data=create_resource(json, db)))
+
+        created_resources_list = []
```
wimo7083: `created_resources = [create_resource(resource, db) for resource in json]`

No need to use hungarian notation, and confuse `res` with resources or response.

app/api/routes.py
```diff
@@ -34,11 +34,21 @@ def resources():
 @requires_body
 @authenticate
 def post_resources():
-    validation_errors = validate_resource(request)
+    json = request.get_json()
```
wimo7083: This feels awkward, having a single route leads to checking for types in multiple places. This makes it harder to encapsulate in tests as your cyclomatic complexity is higher.

app/api/validations.py
```diff
@@ -24,9 +24,19 @@ def missing_json_error():
     return standardize_response(error, status_code=422)
 
 
-def validate_resource(request, id=-1):
+def validate_resource_list(req, rlist):
```
wimo7083: Another case that both list and dict makes it feel awkward. 

Above we call `validate_resource` and then we call `validate_resource_list` but in both cases we call the returned value `resource_errors` 

If we have a list of resources does this mean we could have a list of errors for each resource (or an empty list) and a list of errors for a single resource?

wimo7083:
As we have rate limiting, I'd specify a maximum amount of resources that can be entered at once.

app/api/routes.py
aaron-suarez: Yeah it felt awkward writing it... I'm just not sure how to break it out into two different routes when `POST /resources` already creates a single resource :-(

aaron-suarez:
Regarding the maximum amount of resources, what do you think about 200? The reason I want that amount specifically is because that's about how many I have to add in my next batch and that's about the max I think anyone will ever need to add at once. But reasonably, I don't expect people to add more than like 10 or 20 at the very most through the UI, so maybe I should just break mine down into several requests and have a lower number?

app/api/validations.py
aaron-suarez: Well the beauty of the errors is that each resource will have an error object, which is how the API responds right now for errors. When you try to create a single resource and it fails validation, it gives you an object with all of the errors inside. Now, when you try to create a batch of resources and several of them have errors, you get an error object for each invalid resource. The problem is mapping the error object to the submitted error because the user doesn't submit an identifier. You could use the URL as a unique identifier, except that one way that validation fails is that you may not have a URL at all... so I just did it by index. I really couldn't figure out a better way that evening, and I haven't thought of anything better since 🤷‍♂

app/api/routes.py
aaron-suarez: I feel that something like `POST /resources-list` or `POST /batch-resources` wouldn't be very RESTful..

wimo7083: https://stackoverflow.com/questions/411462/restful-way-to-create-multiple-items-in-one-request


Maybe changing `/resources` to `/resource` and have this be `resource-collection`

We could try something where we formalize what needs to be done to each resource, and then use an iterator on that resource that performs that action. 

But I'm not too sure I like how it makes the single item end up as. 
```python
def iter_resource_json(resources):
    if isinstance(resources, dict):
        yield resources
    elif isinstance(resources, list):
        for resource in resources:
            yield resource
            
            
def create_resources(json):
    
    
    errors = [validate_resource(resource) for resource in iter_resource_json(json['data'])]
    
    if any(error is not None for error in errors):
        return .....
        
    errors = [create_resource(resource) for resource in iter_resource_json(json['data'])]
```

app/api/validations.py
wimo7083: I also wonder what happens if the errors are different for each type. Or if you would want to create the resources that are okay and only return error resources. But that would create a mixed-mode response.

aaron-suarez: I refuse to let the API create some resources but not all. I prefer full idempotency, but that's a tall order. But I think you should be able to fix the problems with your json and then send the same request with those issues fixed

app/api/routes.py
aaron-suarez: I'm actually considering @mega.mac.slice's suggestion of `POST /resources/bulk`. I like keeping it a separate endpoint because otherwise it feels like an awkward overload. Like when you're writing Java and trying to let a method take any type of input so you have like 50 different method signatures for all the permutations? Yeah, gross. I think I'm going to separate it into a different route, so it's really just down to what we name it. I like /resources/bulk if you like it

aaron-suarez: Ended up making it only accept a list.

tests/unit/test_routes.py
```diff
@@ -589,25 +589,6 @@ def test_validate_resource(module_client, module_db, fake_auth_from_oc):
 
     apikey = get_api_key(client)
 
-    # Input validation
-    # IntegrityError
-    response = client.put("/api/v1/resources/2",
```
aaron-suarez: I don't know what this test does. It's a valid request as far as I can tell. In order to generate an IntegrityError, we would need to mock one.

aaron-suarez:
I have a few more tests I want to write, and I still need to update the docs. Then we should be ready.

tests/unit/test_routes.py
```diff
@@ -495,19 +495,33 @@ def test_create_resource(module_client, module_db, fake_auth_from_oc, fake_algol
...
+                            headers={'x-apikey': apikey})
+
+    assert (response.status_code == 422)
+    assert (isinstance(response.json.get("errors")[0].get("too-long"), dict))
+    assert (isinstance(response.json.get("errors")[0].get("too-long").get("message"), str))
+
+
+    # TODO: Add tests for adding multiple resources.
+    #       - Check that all resources submitted get added
+    #       - Check that error handling works correctly (index is correct for given error object)
```
aaron-suarez: I'm going to add the above tests when I get the chance. Whoever reviews, please feel free to suggest more test cases because I'm not sure what else to check on it. Other than the `Too many resources in the list` test, all of the current tests are just refactored from when we were doing a single resource on this endpoint. The edge cases of multiple resources are not explored yet.

app/api/routes.py
```diff
@@ -34,11 +36,17 @@ def resources():
...
+
+    if not isinstance(json, list):
+        return wrong_type("list of resources objects")
```
kylemh: Does this error message get sent to the client? If so, I'd prefer it say something like: 
```"request body must contain valid JSON"```.

Comment (aaron-suarez) [reply app/api/routes.py]: Great question! I'll need to add that to my examples above. The `wrong type` function you see there creates a more human-readable string. So you'll see the object be something like 
```json
{
  "apiVersion": "1.0",
  "data": null,
  "error": 1,
  "errors": [
    {
      "bad-request": {
        "message": "This endpoint accepts a list of resources"
      }
    }
  ],
  "status": "Bad Request",
  "status_code": 400
}
```

aaron-suarez:
I'm mobile right now, but I'll edit the above comment with the exact response when I get back to my computer

app/api/routes.py
aaron-suarez: I'm mobile right now, but I'll edit the above comment with the exact response when I get back to my computer

aaron-suarez: I updated the above comment. This is what it looks like when you send an array to the PUT endpoint:
```json
{
  "apiVersion": "1.0",
  "data": null,
  "error": 1,
  "errors": {
    "invalid-type": {
      "message": "This endpoint accepts a resource object"
    }
  },
  "status_code": 422
}
```

kylemh: Yeah, I'd change the message to what I mentioned `resource object` doesn't mean anything to a client consumer.

app/api/routes.py
```diff
@@ -480,16 +517,21 @@ def create_resource(json, db):
...
     except IntegrityError as e:
         logger.exception(e)
         return utils.standardize_response(status_code=422)
 
+    except (AlgoliaUnreachableHostException, AlgoliaException) as e:
+        if (environ.get("FLASK_ENV") != 'development'):
+            db.session.rollback()
+            logger.exception(e)
+            msg = f"Algolia failed to index new resource '{new_resource.name}'"
+            print(msg)
```
Kandeel4411: Minor thing but I think you forgot a `print(msg)` there, also are we going to make it so that it saves resources that succeeds and return an error with the rest? or will it just fail the whole request and delete all the saved successful ones?

app/api/routes.py
```diff
@@ -465,6 +479,29 @@ def update_resource(id, json, db):
...
+def create_resources(json, db):
+    try:
+        created_resources = []
+
+        for resource in json:
+            res = create_resource(resource, db)
+
+            # Pass along standardized responses generated from create_resource
+            if isinstance(res, tuple):
+                return res
```
Kandeel4411: I am having trouble understanding this part, why are we returning `res` here?

app/static/openapi.yaml
```diff
@@ -560,51 +572,53 @@ paths:
...
+                url: 'https://www.w3schools.com/css/'
         required: true
       responses:
         200:
           description: Created resource
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Resource'
+                $ref: '#/components/schemas/Resource_list'
```
toadzky: why is this underscore delimited instead of camel-case?

app/static/openapi.yaml
```diff
@@ -722,6 +741,96 @@ components:
             message:
               description: Error message
               type: string
+    Error_list:
+      type: object
```
toadzky: `Error_list` is an `object`? why is it not an `array`?

app/static/openapi.yaml
```diff
@@ -722,6 +741,96 @@ components:
...
+      properties:
+        apiVersion:
+          type: string
+          description: API version
+        error:
+          type: integer
+          description: 0 on successful operation, 1 on failure
```
toadzky: why have a flag field for this? shouldn't the rest response code give you this? or the presence of another field that contains the errors? also, if you are going to have a flag field, why would you use a number instead of a boolean?

aaron-suarez: I have no idea, that one was something I added based on your suggestion in slack. If it makes no sense to you, then I guess I misunderstood what you were suggesting in that message. I'm fine taking this boolean flag out altogether because it makes no sense to me why its in there.

app/static/openapi.yaml
aaron-suarez: I was following the existing pattern of `Something_list` that was in there. This file could probably use a little name convention refactoring. I'm not really sure what best practices are

app/api/routes.py
aaron-suarez: If an error was produced, it will be stored in `res` as a tuple. We want to pass that along so `standardize_response` can parse it and pass it along to the client. The assumption is that the only situation in which a resource would be a tuple rather than a dictionary is if there was an error. Though I agree it makes readability a little difficult. Open to suggestions on a better way to do this.

app/api/routes.py
aaron-suarez: The purpose of the message is not to indicate that it's invalid JSON though, so I don't think your suggestion is quite right. The reason you would get this error is if you sent valid JSON that doesn't conform to the schema expected by the API.

Do you think I should say that? `Request body must conform to required schema` and you have to look in the docs to find out what that schema is?

app/static/openapi.yaml
aaron-suarez: I just kind of ran out of names that I could come up with. This object is the exact same as `Error`, except instead of `Error_data` as its data, it has `Error_list_data`. If there is a way in yaml to use the same error object but specify a different `data`, I'd love to use that, but I'm unaware of a way to do that.

That being said, I'm open to other options about what this should be named. `Error_list_object`?

app/api/routes.py
kylemh: > The purpose of the message is not to indicate that it's invalid JSON though

```python
if not isinstance(json, list):
    return wrong_type("list of resources objects")
```

?

aaron-suarez: Yeah, so that's checking if it's a list type, which is the python equivalent of a JSON array. Basically, it will reach this code and pass this check if you send a `{}` instead of a `[]`. You need to send an array of resource objects, rather than a single object.

aaron-suarez: If you send invalid json, this line will catch that and return an error telling you that you sent invalid JSON:
```python
json = request.get_json()
```

kylemh: Sure, so I'd send a message that expresses all of this logic you just explained. Something like: `Expected an array of resource objects, but received an object`

aaron-suarez: Ok, I can make that change. Thanks!

app/api/routes.py
```diff
@@ -480,16 +517,21 @@ def create_resource(json, db):
...
-        logger.exception(e)
-        print(f"Algolia failed to index new resource '{new_resource.name}'")
-
     except IntegrityError as e:
         logger.exception(e)
         return utils.standardize_response(status_code=422)
 
+    except (AlgoliaUnreachableHostException, AlgoliaException) as e:
+        if (environ.get("FLASK_ENV") != 'development'):
+            db.session.rollback()
```
Kandeel4411: Does this work as expected? Specifically does it rollback the committed resources? From searching around it seems its unlikely to do such a thing

Kandeel4411: We could just use `db.session.add(new_resource)` and after all the created resources return error free, then we commit. But I could be wrong so I would love to know more.

Kandeel4411: Or have you yet to implement the slack suggestions?

app/api/routes.py
aaron-suarez: You're right, that should be `logger.`something. I want it to log the name of the resource, might as well make it the same message that we send out to the client

app/api/routes.py
aaron-suarez: Yes, I forgot that you had helped me solve the logic issue associated with this problem. I'll go ahead and update it with your suggestions from slack and you can review that. Thanks for having my back on this

app/static/openapi.yaml
aaron-suarez: Per slack conversation, I'm going to remove this flag altogether from the API

app/api/routes.py
aaron-suarez: Ok now it looks like this in the response:
```json
{
  "apiVersion": "1.0",
  "data": null,
  "error": 1,
  "errors": {
    "invalid-type": {
      "message": "Expected list of resources objects, but found object"
    }
  },
  "status_code": 422
}
```

It can handle if you pass an object, an integer, a number, an array, a boolean or a string. I'm not aware of any other JSON types, so anything else you pass will return an error like so:
```json
{
  "apiVersion": "1.0",
  "data": null,
  "error": 1,
  "errors": [
    {
      "missing-body": {
        "message": "You must provide a valid JSON body to use this endpoint"
      }
    }
  ],
  "status_code": 422
}
```

app/static/openapi.yaml
aaron-suarez: I think there are a few problems with our naming conventions in this file, and I think that existed before I got to it. I'm going to call that out of scope for this PR, and open a separate issue in the repo for it

aaron-suarez: https://github.com/OperationCode/resources_api/issues/259

tests/unit/test_routes.py
```diff
@@ -979,6 +1108,12 @@ def test_method_not_allowed_handler(module_client):
 # Other Routes
 ##########################################
 
+def test_health_check(module_client):
+    response = module_client.get('/healthz')
+    assert (response.status_code == 200)
+    print(response.get_json())
+    assert (response.get_json().get("application").get("status") == "ok")
```
aaron-suarez: I decided to add this in just because I was in the tests and I noticed we didn't have a test for it. It has nothing to do with this PR but I hope that's ok

app/api/routes.py
```diff
@@ -465,31 +486,64 @@ def update_resource(id, json, db):
...
+        except (AlgoliaUnreachableHostException, AlgoliaException) as e:
+            if (environ.get("FLASK_ENV") != 'development'):
+                logger.exception(e)
+
+                # Remove created resources from the db to stay in sync with Algolia
+                for res_id in resource_id_cache:
+                    res = Resource.query.get(res_id)
+                    res.languages.clear()
+                    res = Resource.query.filter_by(id=res_id)
+                    res.delete()
```
aaron-suarez: I know this looks really funky, it's due to my lack of knowledge about flask-sqlalchemy. This was the only way I could successfully get the resource to delete because of the reference to the resource in the `language_identifiers` table. I'm not sure what the difference is between the object returned from `Resource.query.get(res_id)` and the one returned from `Resource.query.filter_by(id=res_id)`.

If someone knows how to delete a resource in a one liner, please comment and I'll update it.

aaron-suarez:
Per slack conversation with @Kandeel4411, we think the network call to delete objects from Algolia would be more expensive than deleting from the db. It's not more simple either way, but at least we were able to find a way to bring the DB calls in the error handling code down from 2 SELECT statements per resource down to one. We can come back to optimize this further if need be

Issue #194, opened 2019-07-21
Add filtering to search endpoint


aaron-suarez:
We might want to be able to search for "python" but also filter based on "paid=false". Right now, I believe our filters work for the `/resources` endpoint but not the search endpoint.

**Examples:**

This currently works to filter for the `/resources` endpoint:
```
http://127.0.0.1:5000/api/v1/resources?language=python&category=books&paid=false
```

Doing that will filter down to just python books that are free. When using the search endpoint, we'll definitely want to be able to filter on paid as well. It might be nice to filter on other information, but we can accept a PR that just filters on paid and iterate on it later.

**Acceptance Criteria:**

A request should be made to the `/search` endpoint like so:
```
http://127.0.0.1:5000/api/v1/search?language=python&category=books&paid=false
```

All the resources returned should contain `paid: false` in their object (i.e., no paid resources are returned).

If you search for `paid=true`, it should return all paid resources.

mc962:
Would this mainly be passing new parameters to what looks like the Algolia service?
And if so, is there any special setup to be able to hit this service locally? (I keep getting an exception for the service being unreachable from my laptop, with the API run in a container, and looking at the .env file, I'm not sure those are real keys)

aaron-suarez:
@mc962 If you'd like to contribute, feel free to add in language and category filters, those are still needed to be able to close this out 😄 I think the hardest part was done by Kandeel because now you can follow the same pattern for the other two and we're good

mc962:
I can do that.

I think part of my issue (and why I didn't do anything for it at the time), was that I couldn't really connect to Algolia to confirm that things were working. Are there some sort dev credentials that we have for this?

aaron-suarez:
We do not. People have just been signing up for a free trial and going from there. I can see about contacting them about dev credentials if you don't think you'll get it done within the 14 days of the trial

PR #219, opened 2019-10-03
Added 'paid' algolia filtering to search endpoint

Kandeel4411:
Hello!
Refers Issue [#194 ](https://github.com/OperationCode/resources_api/issues/194)

I have added 'paid' attribute filtering to the search endpoint using algolia filters, due to not having API key I haven't been able to test it thoroughly.

Possible issue I noticed ( unrelated to this PR ) is that the algolia query API search has a limit of 512 characters where if the user typed a really long query, it would return a 500 error

Let me know if there is anything else I can do, this is my first time 😅

### DIFF
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -243,8 +243,22 @@ def search_results():
     page = request.args.get('page', 0, int)
     page_size = request.args.get('page_size', Config.RESOURCE_PAGINATOR.per_page, int)
 
+    # Fetch the filter params from the url, if they were provided.
+    paid = request.args.get('paid')
+    filters = ''
+
+    # Filter on paid
+    if isinstance(paid, str):
+        paid = paid.lower()
+        # algolia filters boolean attributes with either 0 or 1
+        if paid == 'true':
+            filters += 'paid=1'
+        elif paid == 'false':
+            filters += 'paid=0'
+
     try:
         search_result = index.search(f'{term}', {
+            'filters': filters,
             'page': page,
             'hitsPerPage': page_size
         })
--- a/app/static/openapi.yaml
+++ b/app/static/openapi.yaml
@@ -91,6 +91,12 @@ paths:
           schema:
             type: string
           required: true
+        - in: query
+          name: paid
+          required: false
+          description: Whether the data is paid or not. To search for *free* resources, make a `GET` request to `/search?paid=false`.
+          schema:
+            type: boolean
       responses:
         200:
           description: Search results
@@ -479,12 +485,9 @@ paths:
         - in: query
           name: paid
           required: false
-          description: Whether the resource is paid or not. To search for *FREE* resources, make a `GET` request to `/resources?paid=false`.
+          description: Whether the resource is paid or not. To search for *free* resources, make a `GET` request to `/resources?paid=false`.
           schema:
             type: boolean
-            enum:
-              - true
-              - false
         - in: query
           name: language
           required: false
--- a/tests/unit/test_routes.py
+++ b/tests/unit/test_routes.py
@@ -594,6 +594,22 @@ def test_search(module_client, module_db, fake_auth_from_oc, fake_algolia_save, 
     assert (result.json['data'][0]['url'] == resource.json['data'].get('url'))
 
 
+def test_search_paid_filter(module_client, module_db, fake_auth_from_oc, fake_algolia_save, fake_algolia_search):
+    client = module_client
+
+    # Test on the unpaid resources
+    result = client.get("/api/v1/search?paid=false")
+    assert (result.status_code == 200)
+
+    # Test on the paid resources
+    result = client.get("/api/v1/search?paid=true")
+    assert (result.status_code == 200)
+
+    # Test with invalid paid attribute given ( defaults to all )
+    result = client.get("/api/v1/search?paid=something")
+    assert (result.status_code == 200)
+
+
 def test_algolia_exception_error(module_client, module_db, fake_auth_from_oc, fake_algolia_exception):
     client = module_client
     first_term = random_string()


aaron-suarez:
@Kandeel4411 Thanks for your contribution, I'll review shortly. Regarding your comment about the 512 character limit, can you open an issue in the repo for it and I'll assign you to fixing it? I assume you're willing/able to address it considering that you brought it up 😅

app/api/routes.py
```diff
@@ -243,8 +243,19 @@ def search_results():
...
     page_size = request.args.get('page_size', Config.RESOURCE_PAGINATOR.per_page, int)
 
+    # Fetch the filter params from the url, if they were provided.
+    paid = request.args.get('paid')
+    filters = ''
+
+    # Filter on paid
+    if isinstance(paid, str):
+        # algolia filters boolean attributes with either 0 or 1
+        algolia_bool = 1 if paid.lower() == "true" else 0
```
aaron-suarez: I'm on the fence about this. I feel like if it's not explicitly `"true"` or `"false"` it should return all resources. I might be convinced why this is a better way though, can you discuss why you chose to have it default to false rather than all?

Kandeel4411: I mainly did it this way because the /resource  endpoint filtering also defaulted to false, that way we could expect similar behavior in filtering here as well

aaron-suarez: Oh.. we should probably have the `/resource` filtering default to `all` if an invalid `paid` attribute is given. Can you open up an issue for that? Seems like a beginner friendly one we can allow another contributor to get in on if that's ok

tests/unit/test_routes.py
```diff
@@ -594,6 +594,47 @@ def test_search(module_client, module_db, fake_auth_from_oc, fake_algolia_save,
...
 
+def test_search_paid_filter(module_client, module_db, fake_auth_from_oc, fake_algolia_save, fake_algolia_search):
+    client = module_client
+
+    apikey = get_api_key(client)
+
+    # Create paid and unpaid resource to find it in the search results.
+    paid_resource = client.post("/api/v1/resources",
+                          json=dict(
+                              name=f"Tester1",
```
aaron-suarez: Not that it's a problem, but why have these be format strings rather than just a regular string?

app/api/routes.py
Kandeel4411: Sure thing, do you want me to default to all as well or leave it also with that issue?

tests/unit/test_routes.py
Kandeel4411: Wops, looks like I missed that 😅

tests/unit/test_routes.py
```diff
@@ -594,6 +594,47 @@ def test_search(module_client, module_db, fake_auth_from_oc, fake_algolia_save,
     assert (result.json['data'][0]['url'] == resource.json['data'].get('url'))
 
 
+def test_search_paid_filter(module_client, module_db, fake_auth_from_oc, fake_algolia_save, fake_algolia_search):
+    client = module_client
+
+    apikey = get_api_key(client)
+
+    # Create paid and unpaid resource to find it in the search results.
```
aaron-suarez: Since we're mocking the search results with `fake_algolia_search` there shouldn't need to be anything in the database. We can probably save the cycles and just take this whole thing out.

tests/unit/test_routes.py
```diff
@@ -594,6 +594,47 @@ def test_search(module_client, module_db, fake_auth_from_oc, fake_algolia_save,
...
+                          ),
+                          headers={'x-apikey': apikey}
+                          )
+    
+    # Test on the unpaid resource
+    result = client.get(f"/api/v1/search?paid=false")
+
+    assert (unpaid_resource.status_code == 200)
+    assert (result.status_code == 200)
+    assert (len(result.json["data"]) == 1)
```
aaron-suarez: Since we're mocking the search results with `fake_algolia_search`, it's somewhat pointless to assert on the length of the data. Also it's a problem with tightly coupling the data to the test because if we decided to add a second result to the mock, this would fail even though the endpoint is functioning properly. The only test you have that makes sense to me is calling the endpoint with `paid=true` and asserting a 200 response and then calling the endpoint with `paid=false` and asserting a 200 response.

I'd also like to see some negative unit tests, like calling it with `paid=waffles` or `paid=😊` and asserting that it still returns the results without throwing some kind of 500 error

tests/unit/test_routes.py
aaron-suarez: That being said, if you can make a case for why this is necessary for this test, let me know and I'll hear you out.

app/api/routes.py
aaron-suarez: Yeah, go ahead and change this PR to default to all if `paid` is not a boolean. Thanks

app/static/openapi.yaml
```diff
@@ -91,6 +91,15 @@ paths:
           schema:
             type: string
           required: true
+        - in: query
+          name: paid
+          required: false
+          description: Whether the data is paid or not. To search for *FREE* resources, make a `GET` request to `/search?paid=false`.
```
aaron-suarez: I don't think that the word `free` needs to be both capitalized and italicized, it seems like overkill. Can you choose one and go with that?

app/static/openapi.yaml
```diff
@@ -91,6 +91,15 @@ paths:
...
+          schema:
+            type: boolean
+            enum:
```
aaron-suarez: People should know what a boolean is. We can remove the `enum` part of this. We can also remove it from the `List Resources` portion if you don't mind while you're in there. I didn't notice it before.

PR #221, opened 2019-10-03
add basic status code tests for views

### LINKED ISSUE #207: Test coverage for views
We recently added documentation to the `/` route. We have no tests around those views. It would be nice to verify in the unit tests that they return valid responses so we can be sure that every deploy will have the docs correctly loaded.

Screen Shot 2019-07-29 at 12 16 05 AM

teaglebuilt:
Closes #207 

So beyond checking the status code. I could look into finding a way to create a assertion comparing the static/openapi.yaml file against the downloaded yaml file. I guess i would have to look into how to access the downloaded yaml file without using a hardcoded path in a context manager? What do you think?

### DIFF
--- a/tests/unit/test_routes.py
+++ b/tests/unit/test_routes.py
@@ -4,7 +4,9 @@ from os import environ
 from app.api.auth import blacklist_key
 from app.utils import get_error_code_from_status, msg_map, random_string
 from app.api.validations import MISSING_BODY, MISSING_PARAMS, INVALID_PARAMS
+from app.versioning import LATEST_API_VERSION
 from configs import PaginatorConfig
+from yaml import load
 
 ##########################################
 # Test Routes
@@ -103,6 +105,19 @@ def test_single_resource_out_of_bounds(module_client, module_db):
 def test_get_favicon(module_client):
     response = module_client.get("favicon.ico")
     assert (response.status_code == 200)
+
+
+def test_get_docs(module_client):
+    response = module_client.get("/")
+    assert (response.status_code == 200)
+
+
+def test_open_api_yaml(module_client):
+    response = module_client.get("/openapi.yaml")
+    assert (response.status_code == 200)
+    open_api_yaml = load(response.data)
+    assert (isinstance(open_api_yaml, dict))
+    assert (open_api_yaml.get("info").get("version") == LATEST_API_VERSION)
 
 
 def test_paginator(module_client, module_db):


tests/unit/test_routes.py
```diff
@@ -93,11 +93,24 @@ def test_single_resource_out_of_bounds(module_client, module_db):
...
+    assert (response.status_code == 200)
+
 
 def test_get_favicon(module_client):
     response = module_client.get("favicon.ico")
     assert (response.status_code == 200)
 
+def dir_last_updated(folder):
+    response = module_client.get("favicon.ico")
+    assert (response.status_code == 200)
```
aaron-suarez: This isn't really testing the `dir_last_updated` function. `dir_last_updated` is used in the `/` route, as shows here:
https://github.com/OperationCode/resources_api/blob/4433a9471c25db88f737b29dd38ee57e73457c1b/app/views/routes.py#L16

It's a variable that's available to index.html, as shown by the flask documentation here https://flask.palletsprojects.com/en/1.1.x/api/#flask.render_template (see `context`).

Anyway, the real way to get a behavior-driven test around this function would be to call the function and save the timestamp it returns, then change the file and call the function again and note that the timestamp has changed. However, that doesn't sound too practical for a unit test 🤔

aaron-suarez:
>I could look into finding a way to create a assertion comparing the static/openapi.yaml file against the downloaded yaml file.

That sounds good. How would you compare the files? Hash them and then compare the hashes? I'm not actually sure what the options are for file comparison in python... 

Do you think that's overkill for the unit test? If it's something that interests you, I don't want to stop you, but I wonder if that gives us a lot more confidence than just checking for the 200 response.

A lighter weight option that might give us slightly more confidence than the 200 response is to just parse the file returned into a `yaml.load()` type of function. If that function successfully parses it without throwing an error, then we know that it returned a yaml file and we have a fair amount of confidence that it's the same yaml file because exactly where else would valid yaml come from out of that endpoint?

teaglebuilt:
yeah i guess your right there is no other static yaml file to send. I like the idea of using yaml.load() to make sure that the yaml file was returned from that endpoint.

tests/unit/test_routes.py
teaglebuilt: hey man how do you debug pytest in docker? For example I am trying to understand the module_client fixture and what content it holds in a test. For example module_client.get("")..and what the response object holds?

aaron-suarez: Personally, I debug with print statements, but I should probably get a real debugger one day 😅

teaglebuilt: same, but for example if i want to print a variable in a fixture in conftest? Instead of using make, i run 

```
docker-compose run resources_api py.test /tests/conftest.py
```
so i imagine you might not test with docker since you created all this and runs locally on your box without docker. though in the situation of testing within a container, i was curious how you printed and debugged.

aaron-suarez: When I run `make test` I see the output from print statements even though it uses docker. Sometimes when testing, print statements don't execute because they only print on failures, so I'll do something like `raise exception` right after my print statement so I can see the print. It's not an ideal flow, I think it would be better to get pydebugger or similar working with docker, but I haven't really had the time/motivation for such a task.

If you get something working, please report back (or better yet add something to the README about how to debug tests 😄)

teaglebuilt:
what do you think about also verifying the content type in the headers?

```
def test_open_api_yaml(module_client):
    response = module_client.get("/openapi.yaml")	    response = module_client.get("/openapi.yaml")
    assert (response.status_code == 200)	    assert (response.status_code == 200)

    stream = response.headers[1][1:]

    assert str(stream) == "('application/octet-stream',)"
```

aaron-suarez:
I guess that's fine, but remember that the point of a unit test is to have some assurance that as you make changes, you'll find out if you broke anything. I like the `yaml.load()` idea because if that fails, we know we broke that route. With the content type, maybe it's a good thing if there's a client out there that would break if we returned something other than a stream, but it seems like as long as it's Yaml that gets to them, the content type is less relevant. Can't hurt to leave it in there, but I'd still like to see a test showing that it's valid yaml that is returned from that endpoint

tests/unit/test_routes.py
```diff
@@ -1,6 +1,9 @@
 import pytest, random
 from tests import conftest
 import time
+import os, io
+import yaml
+import re
```
aaron-suarez: I assume this is still a work in progress and that we're only going to keep imports that we actually use at the end of it all?

teaglebuilt: yes i still have not evaluated the yaml file. I wanted to see what you though about evaluating the headers? So with evaluating the yaml file I am trying to find a way to mock the behavior or capture the downloaded file without pulling from an actual path unless its sent to a tempdir

aaron-suarez: See my comment above on evaluating the headers. I'm indifferent. Feel free to leave it in if you like, but it doesn't really verify that the endpoint is working correctly so until we have the yaml.load() this still can't be merged.

PR #226, opened 2019-10-05
default to all when paid param is invalid

### LINKED ISSUE #222: Change 'paid' filtering in resources endpoint
Currently if an invalid paid attribute is given to the `/resource` endpoint it will default to it being `false`, we would like it to default to `all` instead

**Examples**:
This currently defaults to `paid=false` to filter for the /resources endpoint:
```
http://127.0.0.1:5000/api/v1/resources?paid=trueeeeeeee
```
Returns all resource objects with `paid: false`

**Acceptance Criteria**:
```
http://127.0.0.1:5000/api/v1/resources?paid=trueeeeeeee
```
Should return all resources instead ( due to it not being strictly true/false case insensitive )

Pachwenko:
closes #222

Thanks for making an easy issue for me to work on.


app/api/routes.py
```diff
@@ -217,7 +217,7 @@ def get_resources():
         )
 
     # Filter on paid
-    if isinstance(paid, str):
+    if isinstance(paid, str) and paid in ['true', 'false']:
         paidAsBool = paid.lower() == 'true'
```
aaron-suarez: Can we move the `lower()` function call up to the above line so that this still returns all true resources? `/api/v1/resources?paid=TRUE`

We should probably add a unit test around the case sensitivity of all the parameters that we have `lower()` calls like this.

Pachwenko: I updated the behavior and test. Also updated the /update test to also catch that

tests/unit/test_routes.py
```diff
@@ -448,7 +470,7 @@ def test_update_resource(module_client, module_db, fake_auth_from_oc, fake_algol
             languages=["New language"],
             category="New Category",
             url="https://new.url",
-            paid=False,
```
aaron-suarez: The test from line 174 is sufficient, this shouldn't need to change. Can you explain why this is necessary?

Pachwenko: In that route we do the same lower() call and updating this test should cover that scenario. It's not necessary though, and I don't have an issue removing it

aaron-suarez: Ok I see what you did there. I'd like to make sure both scenarios are covered though. Is there another instance where `update_resource` is called in the tests with `paid=False`? If not, can we add a specific test so that it covers both paths?

PR #227, opened 2019-10-05
fix(tests): decoupled hardcoded strings from tests

### LINKED ISSUE #116: Testing should be decoupled from specific response strings

RanadeepPolavarapu:
Resolves #116


aaron-suarez:
@wimo7083 You had a much stronger opinion than I did about how these tests looked. Can you please review this one to see if it solves the problem? I don't think we had clear acceptance criteria when we originally discussed the issue.

aaron-suarez:
@RanadeepPolavarapu I spoke with @wimo7083 on slack and we think the change is fine. If you would please fix your conflicts, I'll go ahead and get this merged

RanadeepPolavarapu:
@aaron-suarez Awesome. Sure, I'll resolve conflicts once I'm home from work.

RanadeepPolavarapu:
@aaron-suarez I have resolved the merge conflicts.

Issue #82, opened 2018-12-28
Fix Code Climate issues


aaron-junot:
We recently added in a Code Climate integration, and it complains about some issues. It's not a huge deal, but it would be nice to address these issues before they get more out of hand.

https://codeclimate.com/github/OperationCode/resources_api/issues

Ideally, I would like to see the `migrations` folder ignored by code climate. I only want our actual source code and tests to be inspected by this tool. Things that are build tools, build artifacts, etc should not be included in this scan.

KPGunner:
After looking at Code Climates documentation I created a .yml to tell Code Climate to ignore folders and files we specify. I added the migrations folder.

wimo7083:
What else is needed for this?

aaron-suarez:
There are now 17 issues. Follow the link to find out more. 
![Screen Shot 2019-04-18 at 10 24 50 PM](https://user-images.githubusercontent.com/2008701/56403830-da2b5e80-6228-11e9-8d36-ed1a48651e03.png)

https://codeclimate.com/github/OperationCode/resources_api/issues

KPGunner:
It doesn't look like any of those would break anything, just mostly complexity issues. Did we want to add the pages to the .yml to ignore them?

aaron-suarez:
I think we can certainly raise the limit for LOC, I think 250 is a bit of a stretch. Alternatively, it might make sense to break up `routes.py` and `test_routes.py` into smaller files, but I don't think they're too big right now, personally.

As for the code complexity issues, if those functions can be refactored, that would be nice, but again, I'm not too worried about it. I'm not even sure how to fix the "too many early returns" thing. Maybe using variables and returning the value at the end? I think that a long `&&` `||` string would be much harder to grok than just "return False"...

How about this: if you want to tweak the yaml file to get rid of all of them, we can review the rules in the PR and make specific decisions point by point. If we want to actually fix the code in any case, we'll do that.

avats-dev:
@aaron-suarez can I work on this? Also, the link given above for code climate issues gives a 404.

aaron-suarez:
@avats-dev I believe that was a 404 because we switched the default branch from `master` to `main` but never updated the default branch in Code Climate. Thanks for letting me know, I made the switch and it appears to be working now. Looks like there's 59 issues identified, so it appears you have your work cut out for you 😆 Thanks for picking this up

avats-dev:
[...]

PR #228, opened 2019-10-06
[refactor] Reduce cognitive complexity from multiple return statements

hydrosquall:
## Overview

- Removes 4 CodeClimate warnings about having many returns, and 1 warning about having high cyclomatic complexity, while preserving the same behavior. This was motivated by #82 .
- Uses a generator so that the test will short-circuit if any of the earlier checks fail.

- See https://codeclimate.com/github/OperationCode/resources_api/app/models.py/source#issue-a49e24d316f36fbe89761497d99f0928 for the warnings of interest.

<a href="https://cl.ly/211ddc5c095e" target="_blank"></a>

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -77,21 +77,20 @@ class Resource(TimestampMixin, db.Model):
         return self.url
 
     def __eq__(self, other):
-        if isinstance(other, Resource):
-            if self.name != other.name:
-                return False
-            if self.url != other.url:
-                return False
-            if self.paid != other.paid:
-                return False
-            if self.notes != other.notes:
-                return False
-            if self.category != other.category:
-                return False
-            if self.languages != other.languages:
-                return False
-            return True
-        return False
+        return isinstance(other, Resource) and all(
+            (
+                getattr(self, attribute)
+                == getattr(other, attribute)
+                for attribute in [
+                    "name",
+                    "url",
+                    "paid",
+                    "notes",
+                    "category",
+                    "languages",
+                ]
+            )
+        )
 
     def __hash__(self):
         return hash(self.url)


app/models.py
```diff
@@ -77,21 +77,20 @@ def key(self):
...
-                return False
-            if self.languages != other.languages:
-                return False
-            return True
-        return False
+        return isinstance(other, Resource) and all(
+            (
+                getattr(self, attribute)
+                == getattr(other, attribute)
+                for attribute in [
```
aaron-suarez: This looks like it might be less efficient because of the nested loop situation (I'm not sure how python actually interprets this, but it looks like a nested loop to me). However, in the interest of fixing the code complexity issue and because this looks more elegant and pythonic, I'm going to try to avoid premature optimization and go with it. Thanks 😄

hydrosquall: This is a good question! The outer loop is a generator, so it should only pull out the attributes from the inner loop 1 at a time. This means that if the first attribute test fails, the other attributes don't get tested as well.

hydrosquall: I'm open to coming back and optimizing this if needed in the future :)

aaron-suarez: I think we're a little early to explore performance testing. Once we do, this might be an area we look at, but it all depends. For now, we need to get the front end coded up and possibly other clients like slackbot hooked into it before we'll have enough users to worry about performance.

Thanks for the offer 😄

hydrosquall: Ah, got it, just saw https://github.com/OperationCode/front-end/issues/661. Is there a public API that can be pinged to test out this resource without using the Slackbot?

aaron-suarez: I'm not sure if you mean to say "Where is the resources API deployed?" But that's the question I'm going to answer. https://resources.operationcode.org has the docs, follow them to use the API

hydrosquall: Ah, despite my awkward phrasing, you figured out the question that I had meant to ask. Thanks!

PR #234, opened 2019-10-14
Added Language & Category filtering to search endpoint

Kandeel4411:
Refers [Issue#194](https://github.com/OperationCode/resources_api/issues/194)
Added Language & category filtering to search endpoint with tests & docs.

Please let me know if there is anything that is needed

Edit:
Tests will fail currently because need actual API Key to be able to set_settings correctly for the attributes that need faceting for algolia, otherwise they should work fine. any thoughts? 

Possible solution:
Add the attributes throught the [dashboard](https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/how-to/filter-arrays/#using-the-dashboard) and remove the set_settings part from the endpoint, that should make the tests work correctly without having to use the API key

PR #236, opened 2019-10-20
Filter by multiple languages on List Resources endpoint

aaron-suarez:
#234 Allows you to filter the search endpoint by multiple languages. This PR is to maintain consistency with both endpoints, but also because it's a useful feature in its own right. It should be noted that Algolia will display results for anything kind of close to either language passed in, whereas the List Resources endpoint makes it a hard secondary filter. So if you try to list Resources for both `Scala` and `Python` for example and there are none that have both of those, it will return no results.

### DIFF
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -165,7 +165,7 @@ def get_resources():
     """
     Gets a paginated list of resources.
 
-    If the URL parameters `language` or `category` are found
+    If the URL parameters `languages` or `category` are found
     in the request, the list will be filtered by these parameters.
 
     The filters are case insensitive.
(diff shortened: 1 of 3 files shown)


aaron-suarez:
You know, that's a really good question. I can see it going both ways depending on how the UI is set up.

I think that you're right though, we should probably make it such that it's the union of all Python resources and all JavaScript resources rather than just the intersection of the two.

It would be nice to have a way to specify whether you'd like the union or the intersection though because I can see a use case for either. Do you have any suggestions on how I might tweak the interface to allow for either/or?

Kandeel4411:
I can't think of any specific way to do it at the moment. if we do it like
```
/resource?languages=python;must&languages=javascript;opt
```
that we require extra parsing and more error handling to check if they sent a valid param. But if we add something like `force` param which could be `bool` for either being a "Must/Optional" it would be alot easier I think. i.e:
```
/resources?languages=python&languages=javascript&force=true
```

aaron-suarez:
I agree, the first is a lot more work than its worth because that's not something that flask can handle natively. The idea of an extra flag to indicate one or the other is good. I'm not sure if I like `force` as the name though because there are several query params accepted by this endpoint and it needs to be clear that it has to do with languages. I'm going to continue to think on it...

aaron-suarez:
After thinking about it, I think that for now, filtering to include all of the results is better than filtering to include only the intersection of the results. I don't particularly like the interfaces I've been able to come up with, and I think that filtering for only the ones that have both can be done on the client side pretty easily. If the front end experiences issues, we can consider adding a flag to allow either type of result.

PR #237, opened 2019-10-20
Fix Update Resource schema

aaron-suarez:
The Update Resource endpoint used the incorrect `Request Body Schema`. This updates it to the correct schema.

Before:
Screen Shot 2019-10-20 at 10 18 06 AM
After:
Screen Shot 2019-10-20 at 10 18 38 AM

### DIFF
--- a/app/static/openapi.yaml
+++ b/app/static/openapi.yaml
@@ -380,7 +380,7 @@ paths:
         content:
           application/json:
             schema:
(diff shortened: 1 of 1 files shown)


aaron-suarez:
@Kandeel4411 If you're good with this PR can you go ahead and merge it? Thanks

Issue #229, opened 2019-10-06
API versioning


aaron-suarez:
**Background**
There have been no breaking changes since the first version was released, so this hasn't come up until now. I believe the best solution to #206 will require a breaking change to the API. That being said, when we make breaking changes to the API we'd like to allow for clients to use the older versions of the API so they have time to upgrade their integration to the new version. So we need a way to let the API determine which version is being requested and then return the response with the correct logic for that particular version.

**Description**
Requests should be able to include an `X-Api-Version` header specifying the version of the API the request is meant for. The API should then execute the logic for that particular API version and return the specific response. 

**Acceptance Criteria**
The API should accept an `X-Api-Version` header.
There should be a standardized way to put certain logic behind a newer API version.
The API should respond with an accurate version number for which logic was executed. Right now I think it just responds with 1.0 no matter what.

Jitsusama:
I'm will to take a stab at this.

Jitsusama:
@aaron-suarez; to clarify things, the APIs you want to version are all defined in _/app/api/route.py_ correct?

aaron-suarez:
Yeah, but more generally, any response from our API is going to have a version along with it, which currently is sort of hard coded here https://github.com/OperationCode/resources_api/blob/master/app/__init__.py#L12

We want that instead to be variable, defaulting to the latest version, but if someone requests a specific API version we should support that.

Jitsusama:
Cool, thanks for that additional information. I'm currently thinking of adding a new decorator that will take in some configuration specifying a mapping of versions to functions that will generate outputs. It would wrap the top level blueprint route and its parameter list will receive the generated output and the route's body can send it out. I'll obviously also have to think about how to handle outputting the version as well.

Does that sound good in theory?

Also, looking at the file you linked, I can't help but notice that you have a base URI for the blueprint that references a version number (_/app/v1_). Will it be confusing to have a separate version number specified in a header from what is already specified in the path? Also, what value does the path specified version number provide if it is not getting used to indicate new APIs? Are these going to refer to 2 separate versions (like one version for the whole solution and another specific to each endpoint?)

aaron-suarez:
For the first part: I'm fine exploring the idea of a decorator, that could probably work. But I don't think it even needs to be that much overhead. I think that it would be more like `if` statements, where a version is mapped to some constant that you check against, so it would look like:
```python
def some_route():
     if api_version_1_1:
        do_something_for_1_1()
    else:
[...]

PR #241, opened 2019-10-24
#229 - Add version specific API logic capabilities

Jitsusama:
This PR (hopefully) addresses #229. It adds a new `@versioned` decorator, which can decorate any function/method that gets called during an active Flask request.

This decorator pulls the version number out of the `X-API-Version` request header and passes it as a new _kwarg_ (named version) to the decorated function. This can be used inside of the function to perform version specific logic. I have applied this to the internal `standardize_response` function that is called for most API response generations and changed it to utilize the version passed as a _kwarg_ instead of the app constant. I've done the same to the health check related endpoints.

Please let me know if I've missed the mark with this PR at all. This also includes tests for all but the health check endpoints as I couldn't find any existing tests around that functionality and I'm not 100% sure how to properly test that they are working.


Jitsusama:
I see that CI build failure. I'm taking a look to see if I can figure out how to get conftest.py to cope with the decorator on that one fixture.

Jitsusama:
Tests are all passing now.

aaron-suarez:
Just doing a quick pass over this, it seems like it will respond with whatever version is specified even if that version doesn't exist. Can we instead do one of two things?

1) Return a 404 and error message like `api_version_not_found` that is similar to other 404s returned from the API
2) Default to the latest version when a version that doesn't exist is specified

I'm thinking number 2 would probably be better (more resilient and less breaking, anyway), but maybe it's better to be explicit that the client messed up so that way they fix it rather than failing silently.

Along these same lines, can we have some negative unit tests around this? What happens when you pass `X-Api-Version: waffles` or `X-Api-Version: -1` or `X-Api-Version: 100000.2`? I want to make sure we explicitly choose how the API responds in these types of situations and that it does the correct thing.

For that reason, I'm ok with coupling the unit tests for a given API version with the version itself. So like, maybe we don't need to test for API version `1.4` until we have a `1.4`. Open to opinions on this.

Jitsusama:
@aaron-suarez; thanks for the comments. I honestly was waiting for that to come up in conversation as it was bothering me as well. I think it would be best to either have a list of valid versions, or somehow define a range. Anything outside of those bounds should either return an error, or fallback to a valid version.

I think a list would be the easiest to handle programmatically, but could get unwieldy over time, unless we considered that the list would get reset in every major version upgrade? A range would be difficult just because of how semver versions work. If you specify a range of 1.1 - 1.6, how would we handle 1.1.1-rc.1 or whatever sort of sub version comes up? I think I’d lean towards passing a range, but only if we agree that the format will always be <MAJOR>.<MINOR>. We could define it in the versioned module alongside the default version. Or, have the largest number in the range become the default as well, so we don’t have to manage as much.

To your other point, the main reason why there are no negative unit tests, is because I wasn’t sure how to handle invalid API versions. It probably would be best to return an error, as anything else is really an undefined condition as far as the user is concerned. They expect a version to be used, but it can’t.

What do you think?

Jitsusama:
Also, I think we can unit test the behaviour quite easily by adding a new optional parameter to the decorator that is the range of valid versions. It can default to the range variable we define. Then, in tests, we can set it to whatever we want.

aaron-suarez:
Huh, I didn't think about passing in a range to the decorator, that's a good idea. Then you can unit test it without having to do something crazy with stubs like I thought we were going to have to do.

I think that you and I are on the same page about errors, so here's what I propose:
1) If the `X-Api-Version` header is a valid version, honor it (happy path)
2) If it's blank, default to the latest version (also happy path, particularly for one-off requests rather than full on integrations)
3) If it's invalid, return an error. I like 404 for this, but I could also see a 422 or something else that makes sense. I think this is nice too because if we completely deprecate a version at some point, we can do so and it doesn't have to be something new we add to the app, we just use the same error that we return for invalid versions by taking it out of the `valid_versions` list/range

To your point about range vs list: we're not really following semver anyway, because semver requires a new major version for breaking changes. With this API, we're going to increment the _minor_ version for breaking changes, so this is really something different than semver (which I don't think would make much sense for API versioning anyway). However, I could see a usecase for something we want to roll out to one particular client.

Let's suppose slackbot has a use case that would completely break the front end, but if we don't have it we can't use it for slackbot. Well, slackbot may end up on a different versioning path in that situation. So maybe there's `1.0, 1.1, 1.2` for the front end but there's `1.1-slack, 1.2-slack` etc for the slack bot. I'm not saying this is likely, but it would be nice to allow for the possibility out of the box rather than limit ourselves to floating point numbers...

That being said, I think maybe we could figure out how to allow for that when the problem arises and just take the easiest path rather than a more difficult one for a _potential_ use case that may never come up. So it's up to you, I'm fine with either, a list would be more extensible but a range serves our purposes for now.

Jitsusama:
@aaron-suarez; I think I have things squared away now. The `@versioned` decorator now takes 2 optional kwargs, one for the list of valid API versions (which will default to a list configured in the _versioned_ module when not specified) and another for whether it will cause a `InvalidApiVersion` exception to be thrown on an invalid API version being requested (defaults to `True`, but I setup the `standardize_response` function to disable this so we don't throw an exception after already honouring the request in the route handler.)

Since we don't have an API that is actually versioned yet, I didn't add this decorator to any existing route handlers. As well, I didn't add an exception handler so that the `InvalidApiVersion` exception will cause a properly formatted response to be generated. I thought that might be best left for a time when it would actually get utilized and thus also be able to be properly tested. What do you think?

app/__init__.py
```diff
@@ -5,15 +5,11 @@
...
 from flask_limiter.util import get_remote_address
+from app.healthcheck import add_health_check
 
 db = SQLAlchemy()
 migrate = Migrate()
 
-API_VERSION = "1.0"
-
-# healthcheck uses API_VERSION so this has to come after it's declared
-from app.healthcheck import add_health_check # noqa
```
aaron-suarez: Side note, I'm super glad we were able to move this and remove the `# noqa` comment 😄

app/utils.py
```diff
@@ -102,7 +104,8 @@ def format_resource_search(hit):
     return formatted
 
 
-def standardize_response(payload={}, status_code=200):
+@versioned(throw_on_invalid=False)
+def standardize_response(payload={}, status_code=200, version=0.0):
```
aaron-suarez: I haven't ran the code yet (I'm about to do that next), so maybe this isn't what it seems. Is it possible though that under some circumstance the API would return `"apiVersion": "0.0"` in a response?

We want to default to the most recent version, not 0.0, so would it be possible to import that value and pop that in as the default value for `version` here?

app/versioning.py
```diff
@@ -0,0 +1,62 @@
+from typing import Callable, List
+
+import flask
+from werkzeug.exceptions import BadRequest
+
+VALID_API_VERSIONS = ['1.0']
```
aaron-suarez: I know we talked about a list before, but thinking more about it, do you think this should be a tuple instead so it isn't mutable?

app/versioning.py
```diff
@@ -0,0 +1,62 @@
...
+                version = flask.request.headers.get('x-api-version', LATEST_API_VERSION)
+                if version not in valid_versions:
+                    if throw_on_invalid:
+                        raise InvalidApiVersion(version)
+                    else:
+                        version = LATEST_API_VERSION
+            else:
+                version = LATEST_API_VERSION
+
+            kwargs['version'] = float(version)
```
aaron-suarez: We talked about keeping the versions a list to support other arbitrary strings such as `1.2-slack`. If you pass that into float you get something like this:
```python
>>> x = '1.2-slack'
>>> float(x)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: could not convert string to float: '1.2-slack'
```

Is there a problem with just keeping it a string? Maybe we could do `str(version)` instead?

tests/unit/test_response_util.py
```diff
@@ -0,0 +1,49 @@
...
+def test_presents_requested_api_version(app, client):
+    @app.route('/endpoint')
+    def endpoint():
+        return standardize_response({})
+
+    api_version = VALID_API_VERSIONS[0]
+    response: Response = client.get(
+        '/endpoint', headers=[('X-API-Version', api_version)])
+
+    assert response.json['apiVersion'] == float(api_version)
```
aaron-suarez: The API version in the current response is a string, I'd like to keep it a string. I think this is where the confusion came from. If you go to https://resources.operationcode.org/api/v1/resources right now, you'll notice that it starts
```json
{"apiVersion":"1.0",
```
We can leave it as a string

app/versioning.py
```diff
@@ -0,0 +1,62 @@
...
+        wrapped_function = valid_versions
+        valid_versions = VALID_API_VERSIONS
+        return decorator(wrapped_function)
+    else:
+        if not valid_versions:
+            valid_versions = VALID_API_VERSIONS
+        return decorator
+
+
+class InvalidApiVersion(BadRequest):
```
aaron-suarez: Instead of returning a `400`, which is pretty generic, do you think we could return a `406 Not Acceptable`?

app/utils.py
Jitsusama: It will never actually use that. That is just present so that I can place it at the end of the parameter list in case you are using positional rags elsewhere in the code base. The decorator will always inject a real value, defaulting to the latest version.

app/versioning.py
Jitsusama: That makes complete sense.

app/versioning.py
Jitsusama: Yeah, I’m groovy with making it back into a string.

app/versioning.py
Jitsusama: I’m used to using 400 for user input errors. I believe not acceptable is specific to the accept header or maybe content type. If you’d like I can take a look of any other standard error could work, but most API users should be familiar with 400 being an input value error.

app/utils.py
aaron-suarez: Can we just leave a default off or would that break things?

app/versioning.py
aaron-suarez: I was always taught that 400 is what you use when there's not a more specific error that you can return. I think you're right about the use for 406. We could consider changing it to an `Accept` header rather than an `x-api-version` header. So you would send something like `Accept: application/1.0` in the request. Maybe that's overkill 🤷‍♂

app/versioning.py
aaron-suarez: Ok, once you make this change I can do another round of testing and see if I find any problems with it. Otherwise, looking good

app/utils.py
Jitsusama: It's required, so that I don't break other parts of code that expects the preceding arguments to maintain their current ordering. That said, I'll change it to point at the latest version to make it look less confusing.

- [ ] set default value of standardize_response version parameter to latest API version

app/versioning.py
Jitsusama: - [ ] change list of API versions to a tuple of API versions

app/versioning.py
Jitsusama: - [ ] change version from being converted to a float to being maintained as a string

tests/unit/test_response_util.py
Jitsusama: - [ ] change float conversion of API version in tests to keep it as a string

app/versioning.py
Jitsusama: I've looked, and honestly, 400 should be the correct response code for this scenario. Everything else has to do with content types or request body content (besides other more strange scenarios). Unless you have a strong reason to do otherwise, I suggest we stick with a 400. I've engaged with a lot of third party APIs, and almost without exception, a 400 is standard for this particular type of scenario.

Jitsusama:
I believe I've addressed the outstanding items.

app/versioning.py
aaron-suarez: Ok good enough for me. As long as it's a 4xx error it's semantically correct, so 400 is fine

tests/unit/test_versioning.py
```diff
@@ -0,0 +1,69 @@
+import pytest
+from flask import Flask, Response
+
+from app.versioning import versioned, LATEST_API_VERSION, InvalidApiVersion
+
+
+@pytest.mark.parametrize('versions', ['1.4', '1.4.1-snapshot'])
```
aaron-suarez: This is interesting. We changed `VALID_API_VERSIONS` to a tuple, but it still works when you pass in a list. I think the API for lists is similar enough to the one for tuples that it doesn't really matter which you pass in. Do you think this is fine or should we change it to a tuple just to provide a code example that shows what you'll actually want to pass in if/when you implement this in the main app logic?

aaron-suarez:
Cool, I looked over the code, I ran it locally, it works just as expected for me. The only thing missing is documentation. Can we please add the `X-Api-Version` header to the documentation? We might want to list which versions are valid, but I'm a little nervous about that going stale if someone forgets to update it. I think we can make that decision once we see it rendered in the docs.

Once we get the docs sorted out, I'll be happy to merge. Thanks so much for this!

tests/unit/test_versioning.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @app.errorhandler(InvalidApiVersion)
+    def handle_exception(exception: InvalidApiVersion):
+        return exception.description, exception.code
+
+    response: Response = client.get(
+        '/endpoint', headers=[('X-API-Version', header_value)])
+
+    response_text = str(response.data, 'utf-8')
+    assert (response_text == f'{header_value} is not a valid API version'
```
aaron-suarez: I like this string check for documenting what this test is actually checking for. However I think it might be contributing to this problem https://github.com/OperationCode/resources_api/issues/116

I'm ok leaving this in for now, but if you can think of a way to decouple the response string from this test then we can go ahead and make that change now

Digbigpig:
`There should be a standardized way to put certain logic behind a newer API version.`

If a request comes in to update_resource with api version 2.5 and we have two versions of update_resource that support v(1.0 - 1.9) and (v2.0 - 2.9), how are we routing the request to the right logic?

Let's add a test for this

Digbigpig:
> 
> 
> `There should be a standardized way to put certain logic behind a newer API version.`
> 
> If a request comes in to update_resource with api version 2.5 and we have two versions of update_resource that support v(1.0 - 1.9) and (v2.0 - 2.9), how are we routing the request to the right logic?
> 
> Let's add a test for this

Never mind, the way it works now doesnt support different versions for the same end point. We can manually route these requests if we run into this scenario

Jitsusama:
Thanks so much @Digbigpig for tackling this for me! I've been meaning to get back to it, but life has been a bit hectic lately.

app/versioning.py
```diff
@@ -0,0 +1,64 @@
+from typing import Callable, Iterable
+
+import flask
+from werkzeug.exceptions import BadRequest
+
+VALID_API_VERSIONS = ('1.0',)
+"""API versions that are valid for a user to request.
+
+NOTE: the first value in this list should be the latest supported version, but the
```
aaron-suarez: Nit: We should change the wording in this comment to say `tuple` not `list`, but not a blocker.

I'll probably go ahead and fix this myself after it's merged.

app/versioning.py
```diff
@@ -0,0 +1,64 @@
...
+            valid_versions = VALID_API_VERSIONS
+        return decorator
+
+
+class InvalidApiVersion(BadRequest):
+    """An invalid API version was requested in the inbound HTTP request."""
+
+    def __init__(self, invalid_version: str):
+        super(BadRequest, self).__init__()
+        self.description = f'{invalid_version} is not a valid API version'
```
aaron-suarez: I wonder if we should sanitize `invalid_version` rather than just sticking it in the string. The thought being that if it was a browser that sent an invalid API version and the client code just displays it on the page, you could inject a cross site scripting `<script>` tag that would then display in the page.

My initial thought is that in general we would expect an invalid version not to make it to production on the front end and that it's also the responsibility of the client to sanitize this string before displaying it in the page.

Something to keep in mind, but I think we can move forward without escaping the string.

tests/unit/test_versioning.py
aaron-suarez: If/when I fix my other nitpick above, I'm going to change this too

PR #242, opened 2019-10-25
README improvements

### LINKED ISSUE #98: README is missing `make` commands
There are a bunch of useful `make` commands that are available via the Makefile, but not all of them are called out. There should definitely be a call out for `make test` which will run all of the tests. Any of the other extremely common make commands should also be called out. There should also be a statement that more commands than just the common ones exist in the makefile, and contributors should go ahead and take a look at it.

Thanks to @SkyelahBay for making me aware of this issue.

dprothero:
- Cleaned up formatting
- Take advantage of ordered list auto-numbering
- Documented sign-up process
- Clarified use of Git Bash for Windows users
- Added example curl command to get apikey
- Documented `make test` and referred to Makefile

Fixes #98

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,19 +118,19 @@ If you see some JSON with a bunch of resources, it worked! If you encounter any 
 
 If you make changes to the models.py or other schemas, you need to run a migration and upgrade again:
 
-```sh
+```
 make migrate
 ```
 
 Before committing, please lint your code:
 
-```sh
+```
 make lint
 ```
 
 And make sure the tests pass:
 
-```sh
+```
 make test
 ```
 
--- a/README.md
+++ b/README.md
@@ -26,17 +26,32 @@ This project provides an API for storing and retrieving learning resources that 
 
 Sometimes these installs can be tricky.  If you get stuck ask for help in the Slack [#oc-python-projects](https://operation-code.slack.com/messages/C7NJLCCMB) channel!
 
+1. If you are not a member of Operation Code, please sign up at https://operationcode.org/join
+    - Provide your email, name, zip code, and a password of your choosing.
+    - You will need this email and password later when creating your API key.
+    - Look for a Slack invite email and join the Slack organization, and then the [#oc-python-projects](https://operation-code.slack.com/messages/C7NJLCCMB) channel.
+
 1. Install [Git](https://git-scm.com/downloads).
-- Choose your OS from the website and follow the prompts.  This installs Git and the Bash Terminal on your machine.
-- Extra: [Git Documentation](https://git-scm.com/doc) for more information on Git.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -26,17 +26,32 @@ This project provides an API for storing and retrieving learning resources that
 
 Sometimes these installs can be tricky.  If you get stuck ask for help in the Slack [#oc-python-projects](https://operation-code.slack.com/messages/C7NJLCCMB) channel!
 
+1. If you are not a member of Operation Code, please sign up at https://operationcode.org/join
+    - Provide your email, name, zip code, and a password of your choosing.
+    - You will need this email and password later when creating your API key.
```
aaron-suarez: Good call 👍

README.md
```diff
@@ -26,17 +26,32 @@ This project provides an API for storing and retrieving learning resources that
...
-6. Run `make all` and then navigate to http://localhost:5000/api/v1/resources
+    - Choose your OS from the website and follow the prompts.  This installs Git and the Bash Terminal on your machine.
+    - Windows users: use the Git Bash Terminal for any of
+      the commands in the remainder of this README.
+      [The Bash Primer](http://www.compciv.org/bash-guide/)
+    - Extra: [Git Documentation](https://git-scm.com/doc) for more information on Git.
+
+1. Fork & Clone
+    - [Fork a repository](https://help.github.com/articles/fork-a-repo/)
+    - Create a local clone of your fork (Step 2 in the document above)
```
aaron-suarez: I know this was here before you got to it, but while we're in here do you think we could either remove or clarify the `Step 2` comment? I think there was a point in time where this was stated in a further step down, but now that it's a part of step 2 it seems really unnecessary to even mention which step is referring to the fork.

aaron-suarez: Also, it looks like it's step 3 now after all

README.md
```diff
@@ -96,6 +128,14 @@ Before committing, please lint your code:
 make lint
 ```
 
+And make sure the tests pass:
+
+```sh
+make test
```
aaron-suarez: Do you know why `test` is syntax highlighted? It's not a problem, but it's kind of weird and I have no idea why it would do that for this one and not the other examples above

aaron-suarez:
Looks great, once we remove that part about "step 2", we're good to go! Thanks so much for this!

README.md
dprothero: Removed it... agreed it's unnecessary.

README.md
dprothero: Looks like some weirdness with GitHub flavored markdown. I removed the `sh` hint from the make commands since it doesn't do anything for us, anyway.

PR #243, opened 2019-10-25
API Key Rotation and Revocation

### LINKED ISSUE #225: API Key Rotation and Revocation
Currently, there is no way for users nor administrators to revoke an API key or rotate the API in the event of compromise. For more information about the use cases, see https://github.com/OperationCode/resources_api/issues/103#issuecomment-480605369.

I'm open to discussion on the best way to handle this. Typically (in my experience) an API is associated with a website that will require you to log in to manage keys. In our case, the API is a completely separate system from the operationcode.org website. We're working towards integrating them via the front end, but something like managing keys from the OC site would be a rather large engineering lift. Therefore, unless someone is extremely dedicated to this and can build out the functionality on the front end, the back end and the API, we're going to have to find a way to provide this functionality using only the API.

Keep in mind that the only way we currently have of authenticating users is via their Operation Code credentials. Currently, there is no concept of an "admin". This might require a slight redesign depending on if someone is clever enough to think of a way to incorporate the functionality in the existing system.

**Acceptance Criteria**

1) Users should be able to ask the API for a new API key (i.e., key rotation). The old key should become invalid when this action is performed.

2) Administrators should be able to perform key rotation for any API key. The interface should allow for rotating it based on email address or API key. This functionality may be safer as a click command rather than through the public API. Doing it as a click command should also solve the problem of us having no admin users, because only admins are able to access the click commands.

3) Administrators should be able to blacklist a user from using the API (i.e., "permanent" revocation). Then, even if that user is still a valid OC user, their API key is no longer valid and they will not be allowed to make another one. Again, this should probably be a click command.

If there are some other aspects of this that I missed, please feel free to comment on this issue with additional criteria. It's encouraged to break this out into multiple PRs for each of the acceptance criteria to avoid having an extremely large code change in one PR.

dprothero:
Fixes #225 

- Added a blacklisted flag and deny entry for any key that's blacklisted. Added tests.
- Click commands to blacklist & reactivate apikeys by apikey or email
- Addded a /api/v1/apikey/rotate endpoint to rotate an apikey
- Click command to rotate a key by apikey or email


app/cli.py
```diff
@@ -162,3 +171,37 @@ def create_tables():
     @algolia.command()
     def reindex():
         reindex_all()
+
+    @apikeys.command()
+    @click.argument('apikey')
+    def blacklist(apikey):
```
aaron-suarez: I [mentioned before](https://github.com/OperationCode/resources_api/issues/225#issuecomment-546477867) the assumption that each user would only ever have one API key. In that world, this method alone would suffice.

In a world in which users have multiple API keys, we're also going to want the ability to blacklist a user altogether. So all of their API keys are invalid, and if they try to request a new one from the `/apikey` or `/new_apikey` endpoint, they would be rejected. Blacklisting a user should only be a click command, not available through the public API.

It doesn't look like you've implemented multiple API keys per user, and if you don't, this is perfectly fine for now. However, I could still see it useful to pass in an email address of a user and blacklist their one API key, so I think that if you're fine with it, I'd like you to add this `blacklist(email)` method as well. You can probably reuse much of the same logic that you have in this method, you just need to pull the key from the database differently.

dprothero: Was thinking the parameter should allow you to use email or key. I will work on that.

For the future, I'm not sure if we need a world with n number of keys, but just 2 keys since the only real use case is key rotation. In that case, we could just add another field like `previous_key` (if populated, then that key also works, if it's `null`, it's ignored). WDYT?

dprothero:
@aaron-suarez I think I've got a good MVP for this issue. Let me know if there are any tweaks you'd like to see made.

app/cli.py
aaron-suarez: Right, probably just two keys, but from what I understand, blacklisting doesn't delete a key, right? So if you rotate your key yearly, then after 2 years you'd have more than 2 API keys. So maybe you only have two active at any given time, but there would still be more than 2 associated with an account so we'd want them all to stay blacklisted.

README.md
```diff
@@ -95,6 +95,12 @@ If you see some JSON with a bunch of resources, it worked! If you encounter any
     - It may be helpful to save this apikey in a secure place like a password manager.
     - Do not hard-code your apikey in your own scripts. Always use an [environment variable](https://www.twilio.com/blog/2017/01/how-to-set-environment-variables.html)
     - You can always re-issue the POST to recover your apikey.
+    - To rotate your apikey (getting a new apikey and making your old key invalid), issue an empty `POST` request to `/api/v1/apikey/rotate` with your apikey included in a header `x-apikey: yourapikey`:
+        ```sh
+        curl -X POST \
+          http://localhost:5000/api/v1/apikey/rotate \
+          -H 'x-apikey: 0a14f702da134390ae43f3639686fe26'
+        ```
```
aaron-suarez: This literally didn't work for me, not sure why...
```
$ curl -X POST   http://127.0.0.1:5000/api/v1/apikey \
   -H 'Content-Type: application/json' \
   -d '{    "email": "aaron.j.suarez1@gmail.com",    "password": "[READACTED]"}'
{
  "apiVersion": "1.0", 
  "data": {
    "apikey": "4e0dec64b7284d04b5c9458d9b58fbd8", 
    "created_at": "2019-07-20 16:23:52", 
    "email": "aaron.j.suarez1@gmail.com", 
    "last_updated": ""
  }, 
  "status": "ok", 
  "status_code": 200
}

$ curl -X POST \
   http://127.0.0.1:5000/api/v1/apikey/rotate \
   -H 'x-apikey: 4e0dec64b7284d04b5c9458d9b58fbd8'
{
  "apiVersion": "1.0", 
  "data": null, 
  "errors": {
    "unauthorized": {
      "message": "You must provide a valid API token in the x-apikey header."
    }
  }, 
  "status": "Unauthorized", 
  "status_code": 401
}
```

Am I doing something wrong?

aaron-suarez:
I'm going to wait on a response before I continue reviewing. I want to be able to play with it and make sure I understand how it's supposed to work before I look at the code again and try to figure out how it does what it does.

If we need to pair on it, I'm totally willing to, just hit me up on slack

README.md
aaron-suarez: If it was blacklisted, then shouldn't it tell me that it's blacklisted?

dprothero: Probably not a good security practice to tell someone why their auth failed. I mean, it's a poor testing experience, but I do think it's best practice not to leak too much information to a potential hacker. It's like saying "your password is incorrect" instead of "your email or password are incorrect" (the former implies the email address is in the system... information we don't need to provide an attacker)

dprothero: So... it works for me, so I wonder if your key was blacklisted? Although, that first step probably shouldn't have worked either (unless you blacklisted it in between those two commands).

```
$ curl -X POST http://127.0.0.1:5000/api/v1/apikey \
>   -H 'Content-Type: application/json' \
>   -d '{"email": "dprothero@twilio.com", "password": "[REDACTED]"}'
{
  "apiVersion": "1.0",
  "data": {
    "apikey": "9a10218e65d540719891f4559cdef5bf",
    "created_at": "2019-10-25 16:24:09",
    "email": "dprothero@twilio.com",
    "last_updated": "2019-10-26 21:01:04"
  },
  "status": "ok",
  "status_code": 200
}

$ curl -X POST \
>   http://127.0.0.1:5000/api/v1/apikey/rotate \
>   -H 'x-apikey: 9a10218e65d540719891f4559cdef5bf'
{
  "apiVersion": "1.0",
  "data": {
    "apikey": "30df75bbdc9f42ada002dd302945fe54",
    "created_at": "2019-10-25 16:24:09",
    "email": "dprothero@twilio.com",
    "last_updated": "2019-10-31 19:57:41"
  },
  "status": "ok",
  "status_code": 200
}
```

dprothero: Ahhh... so I can repro if I blacklist my key. The /apikey endpoint should probably also return a 401 if your key is blacklisted. I will add that.

aaron-suarez: Oh, so I somehow blacklisted my key but it was still returning my blacklisted key? Interesting.. Fwiw, I didn't mean to blacklist my key, but I did have some trouble running `make migrate` and just repopulated my local database from scratch to get around it. Do you think that would have anything to do with my key being blacklisted?

aaron-suarez:
@dprothero Can we go ahead and get documentation for these new endpoints?

It will need to be added to the `paths`, and you'll probably need to create a couple new Request/Response objects https://github.com/OperationCode/resources_api/blob/95d1604a774b4da859660ca03d9ab4b5c42944a2/app/static/openapi.yaml#L30
https://github.com/OperationCode/resources_api/blob/95d1604a774b4da859660ca03d9ab4b5c42944a2/app/static/openapi.yaml#L965

dprothero:
Ok @aaron-suarez , I rebased and updated the openapi spec.

aaron-suarez:
I rebased your branch for you because I didn't want to make you have to do it again

README.md
aaron-suarez: As it turns out, I had to blow away my database and run `make setup` again because `make migrate` failed me for some reason... I'll make an announcement about it in #oc-python-projects when this gets merged in case anyone else runs into that issue

tests/unit/test_routes.py
```diff
@@ -495,171 +564,34 @@ def test_create_resource(module_client, module_db, fake_auth_from_oc, fake_algol
...
+        '/api/v1/resources',
+        json=[dict(notes="Missing Required fields")],
+        headers={'x-apikey': apikey}
+    )
     assert (response.status_code == 422)
-    assert (response.get_json().get("data") == None)
-    assert (response.get_json().get("errors")[0].get("index") == 1)
-
-
-def test_create_resource_wrong_type(module_client, module_db, fake_auth_from_oc, fake_algolia_save):
```
aaron-suarez: I need to put my tests for the multiple resources change back in here, they were lost during the rebase

tests/unit/test_routes.py
```diff
@@ -763,10 +718,11 @@ def test_validate_resource(module_client, module_db, fake_auth_from_oc):
                           )
 
     assert (response.status_code == 422)
-    assert (isinstance(response.json.get('errors').get(INVALID_PARAMS), dict))
-    assert (isinstance(response.json.get('errors').get(INVALID_PARAMS).get('message'), str))
-    assert ("url" in response.json.get('errors').get(INVALID_PARAMS).get("params"))
```
aaron-suarez: We also need to put the decoupled responses back in here, they were lost during the rebase

aaron-suarez:
Ok, I played with it, I love the interface. I read the code, it seems fine. I made a few notes about things that went wrong during my rebase. Once we get those fixed, I'm going to merge. If we discover problems with the code rotation, we can fix them then.

Nice work, I really appreciate this!

aaron-suarez:
Ok, we're rebased, with all the tests, and everything seems to be working properly. In the interest of moving fast and breaking things (and because this has been open since October), I'm going to make the executive decision to go ahead and merge this as is, and if we find some issues we want to fix, we can open them as a github issue and fix them going forward.

Thanks again for your contribution @dprothero!

PR #244, opened 2019-10-25
First round of negative tests

### LINKED ISSUE #78: Need negative unit tests
#70 added a bunch of tests for the routes, but it only tests the happy path (what's supposed to happen). We need "negative" unit tests that test that the app fails gracefully and as expected based on bogus input. In those situations where HTML is returned instead of JSON, or the api returns a 500 instead of a 400 series error, the route should be modified such that it will return the expected/required input. When someone puts in a PR for a negative unit test that requires a change of behavior, we can discuss in that PR what the exact behavior we want ought to be.

Acceptance criteria: 

Each route should have a few negative unit tests with bogus input. The test should verify that JSON is returned (preferably with a helpful error message). The API should never return HTML, so unit tests should verify a valid JSON response. No error should leak any information about our database. The HTTP status code returned should be as specific as possible (e.g., a 422 is favored over a 400 when invalid input is returned).

thinkingserious:
Fixes #78

### DIFF
--- a/tests/unit/test_routes.py
+++ b/tests/unit/test_routes.py
@@ -1,14 +1,9 @@
-import pytest, random
-from tests import conftest
-import time
-from app.models import Resource, Language, Category
-from app.cli import import_resources
 from app.utils import random_string, msg_map, err_map
 from configs import PaginatorConfig
 from datetime import datetime, timedelta
 
 ##########################################
-## Test Routes
+# Test Routes
 ##########################################
 
 
@@ -149,13 +144,13 @@ def test_paid_filter(module_client, module_db):
 
     total_free_resources = response.json['total_count']
 
-    assert all([res.get('paid') == False for res in response.json['data']])
+    assert all([res.get('paid') is False for res in response.json['data']])
 
     response = client.get('api/v1/resources?paid=true')
 
     total_paid_resources = response.json['total_count']
 
-    assert all([res.get('paid') == True for res in response.json['data']])
+    assert all([res.get('paid') is True for res in response.json['data']])
 
     # Check that the number of resources appear correct
     assert (total_paid_resources > 0)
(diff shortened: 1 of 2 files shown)


tests/unit/test_routes.py
```diff
@@ -446,13 +437,65 @@ def test_create_resource(module_client, module_db, fake_auth_from_oc, fake_algol
...
+    notes = "∞"
+    response = create_resource(client,
+                               apikey,
+                               name,
+                               url,
+                               category,
+                               languages,
+                               paid,
+                               notes)
+    assert (response.status_code == 200)
```
aaron-suarez: Does this pass? That's odd. `url` is required so this should fail with a response letting the user know that `url` is required to create a resource 🤔

tests/unit/test_routes.py
```diff
@@ -468,47 +511,75 @@ def test_create_resource(module_client, module_db, fake_auth_from_oc, fake_algol
...
+    notes = "∞"
+    response = update_resource(client,
+                               apikey,
+                               name,
+                               url,
+                               category,
+                               languages,
+                               paid,
+                               notes)
+    assert (response.status_code == 200)
```
aaron-suarez: Note that this one is fine because you don't _have_ to include anything in particular on an update.

aaron-suarez:
Thanks for linting the tests. Would you mind trying to remove the exclusion of the `test` directory from `make lint` and see if it will pass now?
https://github.com/OperationCode/resources_api/blob/e32f5a0f583aaca884265b2930d7143f5ad275b0/Makefile#L61

If it doesn't we can talk about a path forward, but I'm curious because @dprothero brought it up in Slack earlier today and I wonder if this fixed all of the issues.

tests/unit/test_routes.py
```diff
@@ -446,13 +437,65 @@ def test_create_resource(module_client, module_db, fake_auth_from_oc, fake_algol
...
+                               apikey,
+                               name,
+                               url,
+                               category,
+                               languages,
+                               paid,
+                               notes)
+    assert (response.status_code == 422)
+
+    # A String to Big for the DB
```
aaron-suarez: nit: `to` should be `too`

tests/unit/test_routes.py
aaron-suarez: This is why it didn't fail https://github.com/OperationCode/resources_api/blob/d667ac0df19a40f0599e34559708fcf56722716c/tests/unit/test_routes.py#L1004

I didn't realize it had a default category in `create_resource`

aaron-suarez:
I want to get some of these outstanding PRs merged, so I'm going to go ahead and merge it in spite of the spelling error in the comment. Feel free to open a follow on PR to fix that if it bothers you now that you know it's there.

thinkingserious:
Thank you @aaron-suarez!

PR #245, opened 2019-10-25
Fix for search endpoint filtering

Kandeel4411:
Upon further testing (Thanks to chynh!), it appears `request.args.getlist` returns an empty list ( instead of `None` as I originally thought, my bad! ) which lead to `if ininstance(languages,list):` always being true which triggered an AlgoliaException due to filtering being invalid ( `filtering: ()` <- due to languages appending nothing)

Likewise for the `category` filtering; it appends an empty string (i.e: given `/search?category`) which leads to filtering error due to incorrect format

This should hopefully fix that issue.

See #234 for previous implementation and details
See [Filters](https://www.algolia.com/doc/api-reference/api-parameters/filters/) for how algolia filters

### DIFF
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -251,7 +251,7 @@ def search_results():
     filters = []
 
     # Filter on paid
-    if isinstance(paid, str):
+    if paid:
         paid = paid.lower()
         # algolia filters boolean attributes with either 0 or 1
         if paid == 'true':
@@ -260,15 +260,19 @@ def search_results():
             filters.append('paid=0')
 
     # Filter on category
-    if isinstance(category, str):
+    if category:
+        # to not let double quotes conflict with algolia filter format
+        category = category.replace('"', "'")
+
         filters.append(
-            f"category:{category}"
+            f'category: "{category}"'
         )
 
     # Filter on languages
-    if isinstance(languages, list):
+    if languages and '' not in languages:
         for i, _ in enumerate(languages):
-            languages[i] = f"languages:{languages[i]}"
+            # to not let double quotes conflict with algolia filter format
+            languages[i] = 'languages:"{}"'.format(languages[i].replace('"', "'"))
 
         # joining all possible language values to algolia filter query
         filters.append(f"( {' OR '.join(languages)} )")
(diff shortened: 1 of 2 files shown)


Kandeel4411:
This also leads me to think that we might need some regex parsing to disallow using algolia specific filtering characters  i.e:
```
# would lead to filtering error because the generated result would be "languages::python"
/search?languages=:python

# would lead to filtering error because generated result would be "languages: "
/search?languages:%20
```

aaron-suarez:
Oh my! We've got to make sure we have comprehensive negative unit tests for this reason. But is this something that would have been caught by a unit test the way that we're mocking Algolia's response? Perhaps we need to get some integration tests with algolia to run in staging before changes to this endpoint are promoted to production...

aaron-suarez:
Luckily, the changes from #234 haven't made it to production yet, so we can take some time to fix this properly without needing to consider it a hotfix. But we do need to make sure this is merged in before we deploy next

Kandeel4411:
Yeah, we might need to have some integration tests for this endpoint. Because technically when the tests ran on the fake `/search`, it should have thrown a 500 error but it just passed with 200 status code

Kandeel4411:
I believe this should work against all the issues mentioned above, using algolia `'' or ""` string quotes  without the need of any parsing to prevent any filtering mishaps

aaron-suarez:
Ok cool, is this still WIP?

Kandeel4411:
Nope! thought I am not sure of the relevance of the tests

aaron-suarez:
@Kandeel4411 I put it up on staging, and it seems like it's working to me. Can you give it a second set of eyes? Try to break the search endpoint and let me know if you can

https://resources.staging.operationcode.org/api/v1/search

Kandeel4411:
@aaron-suarez thanks, I found a minor silly bug:  in joining the filters it should be: `" AND ".join(filters)` instead of `"AND".join(filters)` otherwise it throws 500 error when combining with paid filter ( it worked with others because string quotes already worked as query delimiter so it registered the AND ). other than this, it looks good to me

aaron-suarez:
No problem, thanks for taking a look. Do you want to open the PR to fix it?

Kandeel4411:
sure

PR #263, opened 2019-12-14
Fix CORS issue

### LINKED ISSUE #250: CORS should allow `*.operationcode.org` so the /resources page can get resources
Per the conversation in this issue https://github.com/OperationCode/front-end/issues/661, the lack of Access Control headers is preventing the front end from getting the resources from the API. We need to add CORS headers, but we should only open it up to the Operation Code frontend.

chynh:
Fixes #250  
- Added a new package to enable CORS
- API routes accept requests from allowed origins(e.g. localhost:3000, operationcode.org, etc. )

### DIFF
--- a/app/api/routes.py
+++ b/app/api/routes.py
@@ -13,7 +13,6 @@ from datetime import datetime
 from prometheus_client import Counter, Summary
 import app.utils as utils
 from os import environ
-from flask_cors import cross_origin
 
 # Metrics
 failures_counter = Counter('my_failures', 'Number of exceptions raised')
@@ -21,18 +20,10 @@ latency_summary = Summary('request_latency_seconds', 'Length of request')
 
 logger = utils.setup_logger('routes_logger')
 
-ALLOWED_ORIGINS = [
-    'http://localhost:3000',
-    r"https:\/\/(www\.)?operationcode\.org",
-    r"https:\/\/(.*\.)?operation-code(-.*)?\.now\.sh"
-]
-
-
 # Routes
 @latency_summary.time()
 @failures_counter.count_exceptions()
 @bp.route('/resources', methods=['GET'], endpoint='get_resources')
-@cross_origin(origins=ALLOWED_ORIGINS)
 def resources():
     return get_resources()
 
@@ -40,7 +31,6 @@ def resources():
 @latency_summary.time()
 @failures_counter.count_exceptions()
 @bp.route('/resources', methods=['POST'], endpoint='create_resources')
-@cross_origin(origins=ALLOWED_ORIGINS[1:])
 @requires_body
 @authenticate
 def post_resources():
@@ -60,7 +50,6 @@ def post_resources():
 @latency_summary.time()
 @failures_counter.count_exceptions()
 @bp.route('/resources/<int:id>', methods=['GET'], endpoint='get_resource')
-@cross_origin(origins=ALLOWED_ORIGINS)
 def resource(id):
     return get_resource(id)
 
@@ -68,7 +57,6 @@ def resource(id):
 @latency_summary.time()
 @failures_counter.count_exceptions()
 @bp.route('/resources/<int:id>', methods=['PUT'], endpoint='update_resource')
-@cross_origin(origins=ALLOWED_ORIGINS[1:])
 @requires_body
 @authenticate
 def put_resource(id):
@@ -88,7 +76,6 @@ def put_resource(id):
 @latency_summary.time()
 @failures_counter.count_exceptions()
 @bp.route('/resources/<int:id>/upvote', methods=['PUT'])
-@cross_origin(origins=ALLOWED_ORIGINS[1:])
 def upvote(id):
     return update_votes(id, 'upvotes')
 
@@ -96,7 +83,6 @@ def upvote(id):
 @latency_summary.time()
 @failures_counter.count_exceptions()
 @bp.route('/resources/<int:id>/downvote', methods=['PUT'])
-@cross_origin(origins=ALLOWED_ORIGINS[1:])
 def downvote(id):
     return update_votes(id, 'downvotes')
 
@@ -104,7 +90,6 @@ def downvote(id):
 @latency_summary.time()
 @failures_counter.count_exceptions()
 @bp.route('/resources/<int:id>/click', methods=['PUT'])
-@cross_origin(origins=ALLOWED_ORIGINS[1:])
 def update_resource_click(id):
     return add_click(id)
 
@@ -112,7 +97,6 @@ def update_resource_click(id):
 @latency_summary.time()
 @failures_counter.count_exceptions()
 @bp.route('/search', methods=['GET'])
-@cross_origin(origins=ALLOWED_ORIGINS)
 def search():
     return search_results()
 
@@ -120,7 +104,6 @@ def search():
 @latency_summary.time()
 @failures_counter.count_exceptions()
 @bp.route('/languages', methods=['GET'])
-@cross_origin(origins=ALLOWED_ORIGINS)
 def languages():
     return get_languages()
 
@@ -128,7 +111,6 @@ def languages():
 @latency_summary.time()
 @failures_counter.count_exceptions()
 @bp.route('/languages/<int:id>', methods=['GET'], endpoint='get_language')
-@cross_origin(origins=ALLOWED_ORIGINS)
 def language(id):
     return get_language(id)
 
@@ -136,7 +118,6 @@ def language(id):
 @latency_summary.time()
 @failures_counter.count_exceptions()
 @bp.route('/categories', methods=['GET'])
-@cross_origin(origins=ALLOWED_ORIGINS)
 def categories():
     return get_categories()
 
@@ -144,7 +125,6 @@ def categories():
 @latency_summary.time()
 @failures_counter.count_exceptions()
 @bp.route('/categories/<int:id>', methods=['GET'], endpoint='get_category')
-@cross_origin(origins=ALLOWED_ORIGINS)
 def category(id):
     return get_category(id)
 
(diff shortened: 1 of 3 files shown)


app/api/routes.py
```diff
@@ -13,26 +13,34 @@
...
+    r"https:\/\/(www\.)?operationcode\.org",
+    r"https:\/\/(.*\.)?operation-code(-.*)?\.now\.sh"
+]
+
 
 # Routes
 @latency_summary.time()
 @failures_counter.count_exceptions()
 @bp.route('/resources', methods=['GET'], endpoint='get_resources')
+@cross_origin(origins=ALLOWED_ORIGINS)
```
aaron-suarez: Do we need CORS headers for the GET endpoints? Or is this just to keep things consistent with all routes? I honestly don't know the answer to this, but I know that there are some resources making their way to the frontend right now (20 of them, right?)

chynh: Correct. The resources are in fact available without CORS headers when you first visit the resources route. However, you won't be able to fetch additional resources from the client-side without CORS headers once you are on that route(e.g. when navigating to pages other than the first resource page). At least, that is my initial thought on how to handle data fetching.

There might be a way to go about it without enabling CORS for GET endpoints and I am looking into Next.js docs to see if it would be possible. Maybe should we circle back to this issue after some research?

aaron-suarez: That sounds reasonable, I'm not super worried about it. Is there any way to test for this locally besides just looking at the headers?

aaron-suarez:
I tried it out locally just now. I don't see all three allowed though, I only see the one for localhost:3000
```bash
$ curl http://localhost:5000/api/v1/resources -I
HTTP/1.0 200 OK
Content-Type: application/json
Content-Length: 8598
Content-Type: application/json
Access-Control-Allow-Origin: http://localhost:3000
Vary: Origin
Server: Werkzeug/0.16.0 Python/3.7.5
Date: Thu, 19 Dec 2019 16:15:34 GMT
```

Is this expected? I would have thought there would be 3 domains allowed for the Access-Control-Allow-Origin header

aaron-suarez:
I'm not a very big fan of having a decorator on every single route for something like this. If you look in the [documentation for Flask-Cors](https://flask-cors.readthedocs.io/en/latest/#resource-specific-cors), it gives this example:
```python
app = Flask(__name__)
cors = CORS(app, resources={r"/api/*": {"origins": "*"}})

@app.route("/api/v1/users")
def list_users():
  return "user example"
```

If we want those same three domains for all routes, wouldn't it make more sense to go ahead and do it like this? But instead of `*` for the origin, we want to narrow it down to the origins we specified.

chynh:
> I'm not a very big fan of having a decorator on every single route for something like this. If you look in the [documentation for Flask-Cors](https://flask-cors.readthedocs.io/en/latest/#resource-specific-cors), it gives this example:
> 
> ```python
> app = Flask(__name__)
> cors = CORS(app, resources={r"/api/*": {"origins": "*"}})
> 
> @app.route("/api/v1/users")
> def list_users():
>   return "user example"
> ```
> 
> If we want those same three domains for all routes, wouldn't it make more sense to go ahead and do it like this? But instead of `*` for the origin, we want to narrow it down to the origins we specified.


Yes, that totally makes more sense and that's how I initially went about it. The only reason why I went with decorating each route is that I thought for PUT, POST routes, localhost should not be included in the list of allowed origins. Right now, I believe the front end is configured to make requests to resources API directly but I suppose we can go with app-level CORS configuration like the above if we modified the front end configuration.

chynh:
> I tried it out locally just now. I don't see all three allowed though, I only see the one for localhost:3000
> 
> ```shell
> $ curl http://localhost:5000/api/v1/resources -I
> HTTP/1.0 200 OK
> Content-Type: application/json
> Content-Length: 8598
> Content-Type: application/json
> Access-Control-Allow-Origin: http://localhost:3000
> Vary: Origin
> Server: Werkzeug/0.16.0 Python/3.7.5
> Date: Thu, 19 Dec 2019 16:15:34 GMT
> ```
> 
> Is this expected? I would have thought there would be 3 domains allowed for the Access-Control-Allow-Origin header

Try ``curl -H "Origin: SPECIFY_ORIGIN_TO_TEST" --verbose \
  http://localhost:5000/api/v1/resources``

If the origin you specified matches a string or regex in the list of allowed origin, you should see the ``Access-Control-Allow-Origin`` in the response header. 

Below is a response that I get:
```shell
$ curl -H "Origin: https://operation-code.operation-code.now.sh" --verbose \
http://localhost:5000/api/v1/resources\?page_size\=3
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 5000 (#0)
> GET /api/v1/resources?page_size=3 HTTP/1.1
> Host: localhost:5000
> User-Agent: curl/7.54.0
> Accept: */*
> Origin: https://operation-code.operation-code.now.sh
>
* HTTP 1.0, assume close after body
< HTTP/1.0 200 OK
< Content-Type: application/json
< Content-Length: 1594
< Content-Type: application/json
< Access-Control-Allow-Origin: https://operation-code.operation-code.now.sh
< Vary: Origin
< Server: Werkzeug/0.16.0 Python/3.7.5
< Date: Thu, 19 Dec 2019 21:39:11 GMT
<
```

I believe Access-Control-Allow-Origin header would only display a single origin, the one where the request came from if it matches one of the strings or regex specified in the origin list.

aaron-suarez:
>I thought for PUT, POST routes, localhost should not be included in the list of allowed origins.

So on the one hand, I think that it would be nice if developers didn't create and update resources on the production API while they're working on the frontend. But I think that the best way to do that is to make the frontend make requests to the staging Resources API when in development mode. And since staging uses the same docker image as production, we would need those headers to allow localhost.

Is there another reason we shouldn't be including localhost in those routes, other than wanting to avoid developers making updates during testing?

chynh:
> Is there another reason we shouldn't be including localhost in those routes, other than wanting to avoid developers making updates during testing?

Nope. That was my sole concern and I agree with your suggestion on using the staging resources API. I'll go ahead and make a new commit including what we have discussed so far later today.

Let me know if there is anything else you want to discuss!

aaron-suarez:
After testing it out, I see how it works and I like it. Other than the fact that I want to DRY this up by having the same headers for all endpoints, I think we're good to go. Thanks for this!

aaron-suarez:
Thanks for your contribution!

chynh:
@aaron-suarez Unfortunately, it seems that this issue has not been completely resolved.
It definitely works locally but I noticed that somehow CORS header is lost when making a call to the production API.

```shell
$ curl -H "Origin: https://www.operationcode.org" --verbose \
https://resources.operationcode.org/api/v1/resources
*   Trying 13.59.34.118...
* TCP_NODELAY set
* Connected to resources.operationcode.org (13.59.34.118) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* Cipher selection: ALL:!EXPORT:!EXPORT40:!EXPORT56:!aNULL:!LOW:!RC4:@STRENGTH
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/cert.pem
  CApath: none
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Client hello (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS change cipher, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES128-GCM-SHA256
* ALPN, server accepted to use h2
* Server certificate:
*  subject: CN=resources.operationcode.org
*  start date: Oct 20 21:02:32 2019 GMT
*  expire date: Jan 18 21:02:32 2020 GMT
*  subjectAltName: host "resources.operationcode.org" matched cert's "resources.operationcode.org"
*  issuer: C=US; O=Let's Encrypt; CN=Let's Encrypt Authority X3
*  SSL certificate verify ok.
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Using Stream ID: 1 (easy handle 0x7fdb58003c00)
> GET /api/v1/resources HTTP/2
> Host: resources.operationcode.org
> User-Agent: curl/7.54.0
> Accept: */*
> Origin: https://www.operationcode.org
>
* Connection state changed (MAX_CONCURRENT_STREAMS updated)!
< HTTP/2 200
< server: nginx/1.13.5
< date: Fri, 20 Dec 2019 17:22:25 GMT
< content-type: application/json
< content-length: 5993
< strict-transport-security: max-age=15724800;
<
```

It seems like we would need to modify Nginx config to fix this but I am totally new to Nginx and not really sure how to go about it.

aaron-suarez:
@chynh I haven't deployed to production yet (we don't have it set up for continuous deployment because I like to test things on staging first). I'm not ready to deploy to production right this second, but try this
```bash
curl -H "Origin: https://operationcode.org" --verbose https://resources.staging.operationcode.org/api/v1/resources
```

chynh:
Got it. I tried the above and it did work on staging API.
Thanks for the clarification!

aaron-suarez:
I'll let you know when we go to prod, I have a breaking change in there for the `POST /resources` endpoint that I want to test a bit more before I deploy to prod

aaron-suarez:
As a matter of fact, feel free to help with that. Check out the documentation for that endpoint here https://resources.staging.operationcode.org/#tag/Create-Resources/paths/~1resources/post

Try creating a few resources with one API call and see what you think as far as whether the docs are clear and they match the behavior you actually see, and/or report if you find anything wonky that needs to be fixed before our front end devs (such as yourself) would find it usable.

chynh:
Sure! 

I wanted to tackle CORS so that I can continue with what I have been working on the resources page front-end. So I am happy to help with anything that can expedite these changes going to prod.

Issue #246, opened 2019-10-26
routes.py and test_routes.py are HUGE


aaron-suarez:
In the past, it didn't bother me that `routes.py` and `test_routes.py` were both over 250 lines of code. However, we've gotten to the point where `routes.py` is close to 500 LOC and `test_routes.py` is almost 1000 (and after #244 is merged, it will be over 1000). It's time to go ahead and break these up into smaller files so we can more easily find what we're looking for.

I'm open to how it should be organized, no strong opinions yet. If I were to do it myself, I'd probably group it similar to [the documentation](https://resources.operationcode.org) where each section was in it's own file. So you'd have `search_routes.py`, `resources_routes.py`, `categories_routes.py` and `languages_routes.py`. Alternatively, there could be new directories for each of these and a `routes.py` in each of them, though I don't know if that makes a lot of sense because I imagine that `routes.py` is all that would go in those directories. The test files would be broken up similarly, though depending on how big each file is, they may need to be broken down further, but we won't know until we go in there and break them up and see what happens. I would hope that each would be less than 250 LOC but I honestly don't know.

Jitsusama:
@aaron-suarez; I'm willing to start tackling this. Would you mind if I do this on a separate PR per organizational boundary? I'm thinking of creating a routes package in place of the routes module, and then having a single module per functional area. The first PR would be groundwork preparing for this, and then the subsequent PRs would be for each functional area and all would be based on the preparatory PR. This should make it easier to review each PR, and, honestly, give me some Hacktoberfest cred 😉

Jitsusama:
Actually, I think I will have a base branch with the setup, and just do a separate PR for each boundary, instead of having a separate setup PR which probably wouldn't be that big anyways.

Jitsusama:
Annndddd, I'm realizing that it's going to be a mess splitting them up into separate PRs. Instead, I think I'll have one PR for routes cleanup and another for route test cleanup. Otherwise, it's probably going to be more confusing to actually go through the PRs, as they will all be dependent on each other (due to modifying the same file over and over again.)

aaron-suarez:
Yeah, I'd much prefer it to be one PR for routes and another PR for the tests. That way, we can verify that moving the routes doesn't break the tests (so do that one first). Then do the tests as a follow up so we can break it down but we can still have confidence that everything is still working fine.

Jitsusama:
Yeah, that's my current plan. Just waiting for feedback before tackling the merge conflicts.

PR #266, opened 2019-12-25
Refactors test_routes as a separate module

Digbigpig:
This refactors the test_routes file as a module. Addresses Issue #246 

Let me know what you think about the file names, whether a test should be moved, or if we should further break down files.

The test_resource_crud.py file (~~holds all CRUD tests rather than GET routes~~) is 550 lines long but doesn't have an unreasonable amount of tests. test_resource_retrieval is just under 200 and I think that is the second biggest file now.

.... I just googled CRUD to realize that the R is for Read, not Remove... So suggestions for a new name are appreciated too

image


tests/unit/test_routes/test_api_key.py
```diff
@@ -0,0 +1,90 @@
...
+
+    response = client.post('api/v1/apikey', json=dict(
+        email="test@example.org",
+        password="supersecurepassword"
+    ))
+
+    assert (response.status_code == 500)
+
+
+def test_get_api_key(module_client, module_db, fake_auth_from_oc):
```
aaron-suarez: Somewhat nitpicky, but can we move this test up to the top? I'd like to have the positive unit tests before the negative ones.

tests/unit/test_routes/test_api_key.py
```diff
@@ -0,0 +1,90 @@
...
+
+def test_rotate_api_key_unauthorized(module_client, module_db):
+    client = module_client
+
+    response = client.post('api/v1/apikey/rotate')
+
+    assert (response.status_code == 401)
+
+
+def test_rotate_api_key(module_client, module_db, fake_auth_from_oc):
```
aaron-suarez: And this positive test should be at the top underneath the other

tests/unit/test_routes/test_categories.py
```diff
@@ -0,0 +1,63 @@
...
+    response = client.get('api/v1/categories')
+
+    for category in response.json['data']:
+        assert (isinstance(category.get('id'), int))
+        assert (isinstance(category.get('name'), str))
+        assert (category.get('name'))
+    assert (response.json['number_of_pages'] is not None)
+
+    # Test trying to get a page of results that doesn't exist
+    too_far = 99999999
```
aaron-suarez: We should break out this negative test into its own function

tests/unit/test_routes/test_categories.py
```diff
@@ -0,0 +1,63 @@
...
+    too_high = 9999
+    response = client.get(f"api/v1/categories/{too_high}", follow_redirects=True)
+    assert (response.status_code == 404)
+
+    # Check GET request to invalid category id - non integer
+    string = 'waffles'
+    response = client.get(f"api/v1/categories/{string}", follow_redirects=True)
+    assert (response.status_code == 404)
+
+    empty_string = ' '
```
aaron-suarez: This is actually a space and not an empty string. We should change this name to `white_space` or something

tests/unit/test_routes/test_languages.py
```diff
@@ -0,0 +1,51 @@
...
+
+    response = client.get('api/v1/languages')
+    for language in response.json['data']:
+        assert (isinstance(language.get('id'), int))
+        assert (isinstance(language.get('name'), str))
+        assert (language.get('name'))
+    assert (response.json['number_of_pages'] is not None)
+
+    # Test trying to get a page of results that doesn't exist
+    too_far = 99999999
```
aaron-suarez: Let's break this out to its own function

tests/unit/test_routes/test_languages.py
```diff
@@ -0,0 +1,51 @@
...
+def test_single_language_out_of_bounds(module_client, module_db):
+    client = module_client
+
+    too_low = 0
+    too_high = 9999
+    response = client.get(f"api/v1/languages/{too_low}", follow_redirects=True)
+
+    assert (response.status_code == 404)
+
+    response = client.get(f"api/v1/languages/{too_high}", follow_redirects=True)
```
aaron-suarez: Let's get some tests with invalid entries like in test_categories.py

And let's add some emojis and weird invalid UTF combinations too, like `%C0%80` or something, let's try to break this thing and make sure we have those same tests for all the `get` routes.

tests/unit/test_routes/test_resource_crud.py
```diff
@@ -0,0 +1,552 @@
+from datetime import datetime
+from app.utils import get_error_code_from_status
+from app.api.validations import MISSING_PARAMS, INVALID_PARAMS, MISSING_BODY
+from .helpers import create_resource, update_resource, get_api_key
+
+
+def test_update_votes(module_client, module_db, fake_algolia_save):
```
aaron-suarez: I think we can make this 3 different functions for the different tests

tests/unit/test_routes/test_resource_crud.py
```diff
@@ -0,0 +1,552 @@
+from datetime import datetime
```
aaron-suarez: This file is still too big. We definitely need to break it up more. I think that the `update` stuff can go in one file, and the `create` stuff should definitely be its own file.

tests/unit/test_routes/test_resource_crud.py
```diff
@@ -0,0 +1,552 @@
...
+        data=bad_json,
+        content_type='application/json',
+        headers={'x-apikey': apikey},
+        follow_redirects=True
+    )
+    assert (response.status_code == 400)
+    assert (isinstance(response.json, dict))
+
+
+def test_validate_resource(module_client, module_db, fake_auth_from_oc):
```
aaron-suarez: There may be a better abstraction for this test rather than grouping the create and update routes together in one validate_resource test... but maybe not. Maybe we have a file called `test_validations`? I'm not sure, open to thoughts, but I don't think this belongs with anything else in this file in its current state. And as I stated above, I'd like to see the update and create tests broken out into two files, I'm not sure how this function will fit into that

tests/unit/test_routes/test_resource_retreival.py
```diff
@@ -0,0 +1,186 @@
+
+from configs import PaginatorConfig
+from datetime import datetime, timedelta
+from app.utils import get_error_code_from_status
+from app.versioning import LATEST_API_VERSION
+from yaml import load
+
+
+# TODO: We need negative unit tests (what happens when bad data is sent)
```
aaron-suarez: We should get rid of this TODO and make sure that we have negative unit tests before we merge

tests/unit/test_routes/test_resource_retreival.py
```diff
@@ -0,0 +1,186 @@
...
+def test_get_resources(module_client, module_db):
```
aaron-suarez: We can probably break this into 3 different functions for the different test cases

tests/unit/test_routes/test_resource_retreival.py
```diff
@@ -0,0 +1,186 @@
...
+    response = client.get(f"api/v1/resources/{too_low}", follow_redirects=True)
+
+    assert (response.status_code == 404)
+
+    response = client.get(f"api/v1/resources/{too_high}", follow_redirects=True)
+
+    assert (response.status_code == 404)
+
+
+def test_get_favicon(module_client):
```
aaron-suarez: This test and the 2 below it don't actually have to do with getting resources. If we need to, we can break this out into its own file as well. Maybe a "junk drawer" type of general tests that are sort of different from the actual routes.. Maybe even throw it in `test_utils.py` which seems to be kind of similar to what I'm describing

tests/unit/test_routes/test_resource_retreival.py
```diff
@@ -0,0 +1,186 @@
...
+    response = module_client.get("favicon.ico")
+    assert (response.status_code == 200)
+
+
+def test_get_docs(module_client):
+    response = module_client.get("/")
+    assert (response.status_code == 200)
```
aaron-suarez: We should add a test that this is actually a valid HTML file

tests/unit/test_routes/test_resource_retreival.py
```diff
@@ -0,0 +1,186 @@
...
+            ('JavaScript' in resource.get('languages'))
+        )
+
+    # Filter by category
+    response = client.get('api/v1/resources?category=Back%20End%20Dev')
+
+    for resource in response.json['data']:
+        assert (resource.get('category') == "Back End Dev")
+
+    # TODO: Filter by updated_after
```
aaron-suarez: Maybe we should go ahead and spend the time to try to figure this one out as well. Or just cut our losses and delete this TODO without that extra test

tests/unit/test_routes/test_searching.py
```diff
@@ -0,0 +1,274 @@
...
+
+def test_search(
+        module_client, module_db, fake_auth_from_oc, fake_algolia_save,
+        fake_algolia_search):
+    client = module_client
+
+    first_term = random_string()
+    apikey = get_api_key(client)
+
+    # Create resource and find it in the search results.
```
aaron-suarez: This is kind of a funky set of tests because it's not actually using Algolia... Like, having an integration test would make more sense than a unit test for this. I think we may want to have a chat about what we're really testing here and take a second look at whether this accomplishes that in the best way. I'm honestly not sure, but I could see this being improved

tests/unit/test_routes/test_utils.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def test_bad_standardize_response(
+        module_client, module_db, unmapped_standardize_response):
+    client = module_client
+    resources = client.get("api/v1/resources")
+
+    assert (resources.status_code == 500)
+    assert (resources.json['errors'] is not None)
+    assert (resources.json['errors']['errors']['server-error'] is not None)
```
aaron-suarez: We should probably assert this is a specific type, rather than just `not None`. I think it's supposed to be a string

tests/unit/test_routes/test_utils.py
```diff
@@ -0,0 +1,112 @@
...
+
+def test_method_not_allowed_handler(module_client):
+    client = module_client
+
+    response = client.patch('api/v1/resources')
+    assert (response.status_code == 405)
+    assert (isinstance(response.json.get('errors')
+            .get(get_error_code_from_status(response.status_code)), dict))
+    assert (isinstance(response.json.get('errors')
+            .get(get_error_code_from_status(response.status_code)).get('message'), str))
```
aaron-suarez: I think we can DRY this up a lot by having some sort of `validate_error` utility function. We could pass in the response and the status code we're expecting (500, 405, 404, etc) and then just have it check the structure of the error response json and that the status code is correct. We repeat ourselves a lot in these negative tests and it would help us cut down on line count and cognitive complexity to have a helper function for these

tests/unit/test_routes/test_resource_crud.py
```diff
@@ -0,0 +1,552 @@
...
+    assert (response.status_code == 422)
+    assert ("boolean" in response.get_json()
+            .get("errors").get("invalid-type").get("message"))
+
+    response = client.put('/api/v1/resources/1',
+                          json="This should be a dict",
+                          headers={'x-apikey': apikey})
+    assert (response.status_code == 422)
+    assert ("string" in response.get_json()
+            .get("errors").get("invalid-type").get("message"))
```
aaron-suarez: We could probably DRY up these `wrong_type` tests with a similar utility function. Pass in the expected type and validate that there's a 422 response and that the expected type is in the response message

tests/unit/test_routes/test_categories.py
Digbigpig: The actual paginator is tested in test_utils. I would suggest just deleting the second half of this test since it adds no value.

tests/unit/test_routes/test_languages.py
Digbigpig: Same with this one.

tests/unit/test_routes/test_resource_retreival.py
```diff
@@ -0,0 +1,186 @@
...
+
+    response = client.get('api/v1/resources')
+
+    # Status should be OK
+    assert (response.status_code == 200)
+
+    resources = response.json
+
+    # Default page size should be specified in PaginatorConfig
+    assert (len(resources['data']) == PaginatorConfig.per_page)
```
Digbigpig: I would move this assert to the Paginator tests in test_utils

app/api/routes/helpers.py
```diff
@@ -11,7 +11,7 @@
 def unauthorized_response():
     message = "The email or password you submitted is incorrect " \
               "or your account is not allowed api access"
-    payload = {'errors': {"invalid-credentials": {"message": message}}}
+    payload = {'errors': {"unauthorized": {"message": message}}}
```
aaron-suarez: The only status code that returns an "invalid-*" response is a 422. For all the rest of the routes, the error is the same as the status, which in the case of 401 is "Unauthorized". I feel this API change is necessary for consistency

tests/unit/test_routes/test_categories.py
aaron-suarez: It probably doesn't add a lot of value because of the paginator tests in test_utils, but I would prefer to keep it just as a warm and fuzzy that every endpoint implements pagination correctly, so I'd like to keep it

app/utils.py
```diff
@@ -120,7 +120,7 @@ def standardize_response(payload={}, status_code=200, version=LATEST_API_VERSION
     elif not data:
         # 500 Error case -- Something went wrong.
         message = msg_map.get(500)
-        resp["errors"] = {'errors': {"server-error": {"message": message}}}
+        resp["errors"] = {"server-error": {"message": message}}
```
aaron-suarez: Fixing these tests to use a standardized utility function to check the responses was really good, because it caught this issue. A frontend would not have been able to parse correctly

tests/conftest.py
```diff
@@ -266,21 +266,10 @@ def algolia_exception(*args):
                  side_effect=algolia_exception)
 
 
-@pytest.fixture(scope='function')
-def fake_validation(mocker):
```
aaron-suarez: I think we were going to use this at some point for something. But since we're not 🤷‍♂

aaron-suarez:
Screen Shot 2019-12-28 at 1 42 04 PM

tests/unit/test_routes/test_searching.py
Digbigpig: Maybe we could start a test automation project to run integration tests on the different parts of OC...

tests/unit/test_routes/test_resource_retreival.py
```diff
@@ -144,5 +148,27 @@ def test_filters(module_client, module_db):
     for resource in response.json['data']:
         assert (resource.get('category') == "Back End Dev")
 
-    # TODO: Filter by updated_after
-    # (Need to figure out how to manually set last_updated and created_at)
+    # Filter by updated_after
+    filter_time = datetime.now() + timedelta(days=-1)
+
+    #   Given I update all resources to be last updated 1 week ago.
```
aaron-suarez: Thanks for including the comments of the walkthrough for my benefit of running through this, but I think when we go to merge, we don't need these comments. The code should speak for itself unless there's something very unclear about _why_ we would be doing a thing.

tests/unit/test_routes/test_resource_retreival.py
```diff
@@ -144,5 +148,27 @@ def test_filters(module_client, module_db):
...
+    set_resource_last_updated()
+    module_db.session.commit()
```
aaron-suarez: I think we should move the commit into `set_resource_last_updated()` unless there's a reason it needs to be here

tests/unit/test_routes/test_resource_retreival.py
```diff
@@ -144,5 +148,27 @@ def test_filters(module_client, module_db):
     for resource in response.json['data']:
         assert (resource.get('category') == "Back End Dev")
 
-    # TODO: Filter by updated_after
-    # (Need to figure out how to manually set last_updated and created_at)
+    # Filter by updated_after
```
aaron-suarez: I think that `test_filters` is testing too many things in one function. Let's break this up into `test_filter_languages`, `test_filter_category` and `test_filter_updated_after`

tests/unit/test_routes/test_resource_create.py
```diff
@@ -0,0 +1,249 @@
...
+                              name="New name",
+                              languages=["New language"],
+                              category="New Category",
+                              url=f"{first_term}",
+                              paid=False,
+                              notes="New notes"
+                          ),
+                          headers={'x-apikey': apikey}
+                          )
+    assert_correct_validation_error(response, ["url"])
```
Digbigpig: @aaron-suarez  This test is expecting for the validation to pass and the commit to the db to fail. When we fail on the db step, the response should be a generic 422 response, not the validation message built by the validation code. So if a "url" validation error exists, we arent testing the scenario that was described in the comment at the top of the test.

This is what I was talking about when I said I couldnt get the mocker to work. I should be mocking the validate resource calls but it looks like it is triggering anyways.

tests/unit/test_routes/test_resource_retreival.py
```diff
@@ -0,0 +1,174 @@
...
+
+
+def test_paid_filter_invalid_paid_parameter(module_client, module_db):
+    client = module_client
+
+    response = client.get('api/v1/resources?paid=na93ns8i1ns')
+
+    assert (True in [res.get('paid') for res in response.json['data']])
+    assert (False in [res.get('paid') for res in response.json['data']])
+
```
Kandeel4411: Is it guaranteed here that there will be an actual mix between  resources paid attribute? I feel like this will fail in the event of the response simply doesn't contain any paid resources for example.

tests/unit/test_routes/test_resource_retreival.py
Kandeel4411: I think so too, we already separated the paid filter, might as well separate those too. Also I just realized that we didn't include the case that there might not be any resources with said languages, perhaps giving a gibberish language and asserting it returns 404? I might be wrong so I would love to know more.

tests/unit/test_routes/test_resource_retreival.py
aaron-suarez: Yeah, I see what you're saying. Luckily, we control the mock data. My thought is that if this test fails and it turns out to the be the data, it should be easy to track down why it happened. If it fails and it's the code not doing the right thing, that's the value of this test

tests/unit/test_routes/test_resource_retreival.py
aaron-suarez: Good call, we should add that test. Same with `/api/v1/resources?category=gibberish` returning a 404

tests/unit/test_routes/test_resource_create.py
aaron-suarez: Yeah, I spent about 2 hours working on this and I can't understand why it will mock the validate_resource call on the `POST` request but not on the `PUT` request...

I'm not really sure what to do to be honest, but I'm tired of messing with it to be honest.

tests/unit/test_routes/test_searching.py
```diff
@@ -0,0 +1,212 @@
...
+                          ),
+                          headers={'x-apikey': apikey}
+                          )
+
+    assert (response.status_code == 500)
+    assert ("Algolia" in response.get_json().get("errors")[0].get("algolia-failed").get(
+        "message"))
+
+    # Set it back to development for other tests
+    environ["FLASK_ENV"] = "development"
```
Kandeel4411: Isn't this test the same as the previous test? Should we separate the body into a separate function and just make each test call it with different parameters? Or would that be overdoing it?

aaron-suarez: Yes, it's the same test except the mocked exception is `fake_algolia_unreachable_host` instead of `fake_algolia_exception` (which is two different errors we want to ensure we're catching in these 4 situations).

If you want to tighten this up and combine them somehow or break the logic out into a function, feel free, just so long as it's still readable

tests/unit/test_routes/test_resource_create.py
Digbigpig: Oh I'm glad I wasn't doing something wrong at least.  I'm okay with this for now. It isn't a very likely scenario anyway.

aaron-suarez:
I documented the remaining problems in a new issue. I'm not going to touch that issue. If someone comes along and wants to mess with them, great. If not, that issue serves more as documentation so if someone comes along they can have some background as to why these things remain as problems. I think we're good to merge this. Please open additional PRs or issues for remaining problems that were not addressed in this PR.

PR #247, opened 2019-10-26
#246 - Break Apart routes Module Into Package

Jitsusama:
This PR, as per the request in #246, breaks apart the routes module into a new package, where separate modules organized by functional area.

PR #267, opened 2019-12-25
#246 - Break Apart routes Module Into Package - Try 2

Jitsusama:
This is a retry on #247, hopefully properly taking into account the current state of master. I tried rebasing #247, but it was too big of a deal to accomplish.


Jitsusama:
Added #noqa line to try and resolve flake8 error, but no joy. @aaron-suarez; can you take a look? I have no clue why Flake is ignoring that comment.

app/api/routes/__init__.py
```diff
@@ -0,0 +1,3 @@
+from app.api.routes import (
+    api_key, categories, languages, resource_creation, resource_modification,
+    resource_retrieval, search)  # noqa
```
aaron-suarez: To make the linter pass, you need the `# noqa` comment be up on line 1

app/api/routes/helpers.py
```diff
@@ -0,0 +1,32 @@
...
+
+from app import utils as utils
+from app.models import Category, Language
+
+logger = utils.setup_logger('routes_logger')
+latency_summary = Summary('request_latency_seconds', 'Length of request')
+failures_counter = Counter('my_failures', 'Number of exceptions raised')
+
+
+def _unauthorized_response():
```
aaron-suarez: Now that we have a separate file for `helpers`, I don't think this needs to start with an underscore anymore. Either that, or `get_attributes` should be `_get_attributes`. I'd like these to be uniform, I don't care which we go with.

aaron-suarez: I checked and the only time we have a `_` in front of functions is for the ones that start with two (like `__eq__`) and the time that we `import db as _db`. I don't remember why we don't just import it as `db`, but I remember there was a reason. Something to do with having `db` in the same scope at some point, I'm not sure.

Either way, this project doesn't really use the convention of `_function_name` for helper functions, so I'm going to rename this as `unauthorized_response()`.

If someone wants to go in and create a convention of doing that for helper functions, they're welcome to, but they need to do it for all of them rather than a random one-off helper function.

aaron-suarez:
This also fixes #246

PR #272, opened 2020-01-10
Allow only one vote per user on resources

### LINKED ISSUE #211: Voting for resources should be limited to one per user
Currently, you can upvote or downvote a resource over and over again. It should be limited to one vote per user. If the user has upvoted a resource and then downvotes the same resource, the upvote should clear out and be replaced by the downvote (and vice versa).

For the moment, users can be tracked by their API credentials, so all of this should be determined by the API key provided.

Acceptance criteria:
- The `/upvote` and `/downvote` endpoints should need authentication
- Votes should be restricted to one per API key
- Votes from a certain user should be able to be cleared out by sending an additional request to either of the vote endpoints. If it is the same vote endpoint, it should simply be cleared out. If it is the opposite vote endpoint, it should clear out the existing vote and update the vote.

chynh:
**Changes**

- Added ``VoteInformation`` model that stores a user's vote information on a resource.
- Modified/refactored API routes to require authentication when voting and implement the one-vote-per-user requirement.
- Modified/Added tests to accommodate model and routes change.

Stopped short of adding more tests to get an opinion on the changes so far. Let me know what you think.

Fixes #211


aaron-suarez:
@chynh I don't see it on here, but I previously recall discussing how this leaks the API keys of all users that have voted for a resource. Were you planning on coming back to this to redesign in a more secure way?

chynh:
Not sure what happened there. But I just pushed it up again so let me know if you still don't see it. I believe with the security concern has been addressed with the redesign.

app/api/routes/resource_modification.py
```diff
@@ -96,16 +96,12 @@ def update_resource(id, json, db):
...
-
-@latency_summary.time()
-@failures_counter.count_exceptions()
-@bp.route('/resources/<int:id>/downvote', methods=['PUT'])
-def downvote(id):
-    return update_votes(id, 'downvotes')
+@bp.route('/resources/<int:id>/<string:vote_direction>', methods=['PUT'])
+@authenticate
+def change_votes(id, vote_direction):
+    if vote_direction not in ['upvote', 'downvote']:
```
aaron-suarez: I love this. Much better than making this two separate endpoints. ❤️👍

aaron-suarez:
I ran this and I only get 400 errors:
```
172.20.0.1 - - [08/Mar/2020 16:41:55] "PUT /api/v1/resources/2/upvote HTTP/1.1" 400 -
172.20.0.1 - - [08/Mar/2020 16:40:47] "PUT /api/v1/resources/2/downvote HTTP/1.1" 400 -
```
Additionally, I don't get the 404 when it's something other than upvote/downvote, I still get a 400
```
172.20.0.1 - - [08/Mar/2020 16:40:19] "PUT /api/v1/resources/2/bogus HTTP/1.1" 400 -
```
Does this code actually work for you?

chynh:
```192.168.176.1 - - [08/Mar/2020 18:59:07] "PUT /api/v1/resources/5/upvote HTTP/1.1" 200 -
192.168.176.1 - - [08/Mar/2020 18:59:08] "PUT /api/v1/resources/5/upvote HTTP/1.1" 200 -
192.168.176.1 - - [08/Mar/2020 18:59:13] "PUT /api/v1/resources/5/downvote HTTP/1.1" 200 -
192.168.176.1 - - [08/Mar/2020 18:59:13] "PUT /api/v1/resources/5/downvote HTTP/1.1" 200 -
192.168.176.1 - - [08/Mar/2020 18:59:26] "PUT /api/v1/resources/5/down HTTP/1.1" 302 -
192.168.176.1 - - [08/Mar/2020 18:59:26] "GET /404 HTTP/1.1" 404 -
```

hmm... not sure what's going on there. Above are the results that I got. I did run into the issue of not getting 404 but other than that, hitting upvote and downvote should work. 

I did fix the 404 issue and pushed it up again. Let me know if it still does not work for you.

aaron-suarez:
@chynh I realized that the 400 error is coming from the fact that my `PUT` body was empty. I'll have to double check the documentation but I don't think we need to require a `{}` in the body, do we? What are you putting in your body?

aaron-suarez:
Ok, I looked at the docs, we don't have docs for this endpoint yet 😅 Can you go ahead and update the [openapi file](https://github.com/OperationCode/resources_api/blob/master/app/static/openapi.yaml) to add docs for these endpoints? We'll need one section for `/upvote` and another for `/downvote`

chynh:
> I'll have to double check the documentation but I don't think we need to require a `{}` in the body, do we? What are you putting in your body?

Ah okay. And yes we don't need the body when making this request.

> Can you go ahead and update the [openapi file](https://github.com/OperationCode/resources_api/blob/master/app/static/openapi.yaml) to add docs for these endpoints?

Got it. I never wrote openapi docs before but looking at the file it seems pretty straight forward. I'll push it up again after adding those sections.

aaron-suarez:
It's possible that Flask just automatically tries to parse a body on `PUT`s, or maybe it was the client I was using was trying to send some sort of whitespace... If there's a way to configure it to ignore the body altogether that would be nice...

app/api/auth.py
```diff
@@ -116,6 +117,9 @@ def log_request(request, key):
     method = request.method
     path = request.path
     user = key.email
-    payload = request.json
+    try:
+        payload = request.get_json()
+    except HTTPException:
+        payload = None
```
aaron-suarez: Looking at the [documentation for `request.get_json()`](https://flask.palletsprojects.com/en/1.1.x/api/#flask.Request.get_json), I would think that we can go ahead and just pass `silent=True` to it rather than do this whole try/except thing with an additional import.

```python
    user = key.email
    payload = request.get_json(silent=True)
    logger = create_logger if method == "POST" else update_logger
```

This works for me locally

chynh:
> It's possible that Flask just automatically tries to parse a body on `PUT`s, or maybe it was the client I was using was trying to send some sort of whitespace... If there's a way to configure it to ignore the body altogether that would be nice...

I figured out that 400 error was coming from ``log_request`` trying to parse json payload when no payload/invalid json is provided. So I added error handling for those cases. Also, added openapi docs for ``upvote`` and ``downvote``

app/api/auth.py
chynh: Ah okay. this is definitely cleaner. I'll just change it as you suggested.

app/static/openapi.yaml
```diff
@@ -517,6 +517,94 @@ paths:
         404:
           $ref: '#/components/responses/NotFound'
 
+  /resources/{id}/upvote:
+    put:
+      summary: Upvote Resource - Increment Number of Upvotes
+      description: |
+        Increments the `upvotes` property of a resource. If the user has already upvoted the resource, upvotes property will decrement since a user is limited to only one vote per resource.
+      tags:
+        - Upvote Resource
```
aaron-suarez: This should be `- Update Resource`. The section for upvoting docs doesn't render because of this. Once this change is made, I think we're good to go, but I'm going to give it a quick once over again before we merge

app/static/openapi.yaml
```diff
@@ -517,6 +517,94 @@ paths:
         404:
           $ref: '#/components/responses/NotFound'
 
+  /resources/{id}/upvote:
+    put:
+      summary: Upvote Resource - Increment Number of Upvotes
```
aaron-suarez: nit: I think we can remove the `Increment number of Upvotes` and just make it `Upvote Resource` and `Downvote Resource`

app/static/openapi.yaml
```diff
@@ -517,6 +517,94 @@ paths:
...
+      description: |
+        Increments the `upvotes` property of a resource. If the user has already upvoted the resource, upvotes property will decrement since a user is limited to only one vote per resource.
```
aaron-suarez: Do you think we should also go ahead and call out that if a user has upvoted a resource and they use the `/downvote` endpoint, it will remove their upvote and replace it with a downvote? Or is that pretty obvious by saying that "a user is limited to one vote per resource"?

chynh: Yeah. I think it makes more sense to mention that. I'll update the descriptions for upvote and downvote.

app/models.py
```diff
@@ -193,3 +195,21 @@ def __repr__(self):
...
+    voter_apikey = db.Column(db.String, db.ForeignKey('key.apikey'), primary_key=True)
+    resource_id = db.Column(db.Integer, db.ForeignKey('resource.id'), primary_key=True)
+    current_direction = db.Column(db.String, nullable=False)
+    resource = db.relationship('Resource', back_populates='voters')
+    voter = db.relationship('Key', back_populates='voted_resources')
+
+    @property
+    def serialize(self):
+        return {
+                'vote_apikey': self.voter_apikey,
```
aaron-suarez: Maybe this is paranoia, but I feel like we shouldn't be serializing the API key itself because if this were to be logged anywhere, we don't want to log API keys in plaintext... but as far as I can tell, we're not using this serialize function anywhere at this time, right?

I think anyone choosing to use the serialize function in the future would look at it and know what date it contains and therefore be careful about what they do with that serialized data. I don't know, I'm just not sure.

Do you have a strong opinion on this one way or the other? If not, maybe we should just take it out and if we ever need the API key itself during serialization we can add it back in at that time?

aaron-suarez: The API key is also logged in the `voter` property, which I'm pretty sure we need.. The serialized info for the usage of this in the test looks like this:
```
{'vote_apikey': '1234abcd', 'resource_id': 1, 'current_direction': 'upvote', 'voter': <Key email=test@example.org apikey=1234abcd>, 'resource': <Resource 
        Name: name
        Languages: [<Language language>]
        Category: <Category Category>
        URL: https://resource.url
>}
```
So maybe we just leave both in there and it's fine

aaron-suarez: Or maybe we should take the API key out of the serialize function for API key and just log the email address of the user? Why would we ever want to serialize the key itself? That just seems like a data leak waiting to happen...

I keep going back and forth, I'm open to suggestion on this

chynh: Okay. It's been a while since I worked on it and I can't remember the reason behind the decision to add serialize function...But I understand the concern here. Let me know if this could be an issue later on and I can circle back to it.

aaron-suarez: Well the reason is because if you do something like `print(vote_info)`, you get something rather unhelpful like
```
<VoteInformation (transient 139740452757904)>
```

Since `transient 139740452757904` tells me nothing about the specific VoteInformation I'm logging, I instead use the `.serialize` function to create a dictionary out of it.

In the case of `Resources` and `Languages`, it also provides a simple way to turn the model into a dictionary so we can return all that data in responses. I think we use that same pattern with the API key itself, when we give the API key to the user via the `/apikey` endpoint.

So, come to think of it, since we're not returning the data within a VoteInformation in a response, I think we probably don't need the serialize function at all. Let's just go ahead and take it out for now and if we need it for some reason in the future we can add it back in. Does that sound good?

chynh: Yeah. Sounds good. Then I'll probably have to take out test for VoteInformation Model as well, right?

aaron-suarez: You don't have to take it out, you just have to assert on each individual property rather than on the serialize

Issue #279, opened 2020-03-04
Investigate issue with Rate Limiting


aaron-suarez:
The rate limiting logic has caused the healthchecks from EKS to cause a denial of service to all clients of the API. More investigation is needed, but it looks like we need two fixes:
1) Whitelist the IP address spaces of EKS so it doesn't get rate limited
2) Ensure that when rate limiting happens, only the problematic client gets rate limited, not all clients everywhere. The whole point of rate limiting is to prevent denial of service, not cause it 😅

irvingpop:
Let's just not have a rate limit for `/healthz`, I think that would solve the operational issues

aaron-suarez:
Alright, to remove rate limiting, it looks like we need to add `@limiter.exempt` to that particular endpoint. Then we need to figure out whether rate limiting applies to all clients and make that not the case

PR #271, opened 2020-01-10
Add unpinned dependencies to fix #254

aaron-suarez:
These 4 dependencies are used but were unpinned.

PR #280, opened 2020-03-04
To fix the rate limiting for /healthz, reorganize the app/__init__ a bunch to support decorators

irvingpop:
This PR had to do a lot, sorry about the diff:
* reorganize `app/__init__.py` to enable decorators and merge healthcheck.py into it
* Update run.py and conftest.py to work with the reorganized app/__init__
* Fix a werkzeug import error for DispatcherMiddleware in run.py

This passes all local testing, `make test` and `make lint`

Signed-off-by: Irving Popovetsky <irving@popovetsky.com>


irvingpop:
paging @aaron-suarez @AllenAnthes for review

requirements.txt
```diff
@@ -1,16 +1,16 @@
...
-pytest-mock==2.0.0
-pytest-cov==2.8.1
-pyyaml==5.3
-requests==2.22.0
-sqlalchemy==1.3.13
-SQLAlchemy-Utils==0.36.1
-prometheus_client==0.7.1
-uWSGI==2.0.18
-py-healthcheck==1.9.0
-bandit==1.5.1
```
aaron-suarez: We want to keep out dependencies pinned. The problem we're seeing with Werkzueg in production is because that package _isn't_ pinned. Please take this change out

irvingpop: want me to revert this entirely (security vulns!) or just bump all the pins?

kylemh: Security vuln resolution sounds like a separate PR

aaron-suarez: Yeah, revert it for now. Feel free to review #271 and check that everything is up to par security wise, and that everything had been included

irvingpop: y'all are some party poopers.  reverted.

aaron-suarez:
fixes #279

aaron-suarez:
I don't remember how it works, so I may not be able to tell you how to fix it. But in this PR, it's missing the `/metrics` endpoint. On master (once you replace the import for DispatcherMiddleware) this is what currently happens:
```
$ curl localhost:5000/metrics
# HELP python_gc_objects_collected_total Objects collected during gc
# TYPE python_gc_objects_collected_total counter
python_gc_objects_collected_total{generation="0"} 711.0
python_gc_objects_collected_total{generation="1"} 282.0
python_gc_objects_collected_total{generation="2"} 13.0
# HELP python_gc_objects_uncollectable_total Uncollectable object found during GC
# TYPE python_gc_objects_uncollectable_total counter
python_gc_objects_uncollectable_total{generation="0"} 0.0
python_gc_objects_uncollectable_total{generation="1"} 0.0
python_gc_objects_uncollectable_total{generation="2"} 0.0
# HELP python_gc_collections_total Number of times this generation was collected
# TYPE python_gc_collections_total counter
python_gc_collections_total{generation="0"} 196.0
python_gc_collections_total{generation="1"} 17.0
python_gc_collections_total{generation="2"} 1.0
# HELP python_info Python platform information
# TYPE python_info gauge
python_info{implementation="CPython",major="3",minor="7",patchlevel="6",version="3.7.6"} 1.0
# HELP process_virtual_memory_bytes Virtual memory size in bytes.
# TYPE process_virtual_memory_bytes gauge
process_virtual_memory_bytes 1.82902784e+08
# HELP process_resident_memory_bytes Resident memory size in bytes.
# TYPE process_resident_memory_bytes gauge
process_resident_memory_bytes 5.1183616e+07
# HELP process_start_time_seconds Start time of the process since unix epoch in seconds.
# TYPE process_start_time_seconds gauge
process_start_time_seconds 1.58338375218e+09
# HELP process_cpu_seconds_total Total user and system CPU time spent in seconds.
# TYPE process_cpu_seconds_total counter
process_cpu_seconds_total 0.0
# HELP process_open_fds Number of open file descriptors.
# TYPE process_open_fds gauge
process_open_fds 9.0
# HELP process_max_fds Maximum number of open file descriptors.
# TYPE process_max_fds gauge
process_max_fds 1.048576e+06
# HELP request_latency_seconds Length of request
# TYPE request_latency_seconds summary
request_latency_seconds_count 0.0
request_latency_seconds_sum 0.0
# TYPE request_latency_seconds_created gauge
request_latency_seconds_created 1.5833837524447525e+09
# HELP my_failures_total Number of exceptions raised
# TYPE my_failures_total counter
my_failures_total 0.0
# TYPE my_failures_created gauge
my_failures_created 1.5833837524447982e+09
```

On your branch I get this

```
$ curl localhost:5000/metrics
{
  "apiVersion": "1.0", 
  "data": null, 
  "errors": {
    "not-found": {
      "message": "The resource you requested does not exist."
    }
  }, 
  "status": "Not Found", 
  "status_code": 404
}
```

app/__init__.py
```diff
@@ -14,30 +18,50 @@
...
+migrate.init_app(app, db)
 
-    db.init_app(app)
-    migrate.init_app(app, db)
+from app.api import bp as api_bp # noqa
+app.register_blueprint(api_bp, url_prefix='/api/v1')
 
-    from app.api import bp as api_bp
-    app.register_blueprint(api_bp, url_prefix='/api/v1')
+from app.views import bp as view_bp # noqa
```
aaron-suarez: I'm sure it's necessary, but why do we need the `# noqa` here? What's the rule that it's breaking that we need to ignore? Can we fix it somehow rather than ignoring it?

Totally not a blocker, but I want to understand because I feel like it might be adding tech debt

app/__init__.py
```diff
@@ -14,30 +18,50 @@
...
-    app.register_blueprint(view_bp)
+from app.errors import bp as error_bp # noqa
+app.register_blueprint(error_bp)
 
-    from app.errors import bp as error_bp
-    app.register_blueprint(error_bp)
 
-    add_health_check(app)
+@app.route("/healthz")
+@limiter.exempt
```
aaron-suarez: Can we add a unit test for the `/healthz` endpoint that makes many requests and verifies that it's not rate limited anymore? I saw that it works I just want to be able to verify that future changes don't break this again

app/__init__.py
```diff
@@ -14,30 +18,50 @@
...
+def healthz():
+    health = HealthCheck()
+    health.add_section("application", application_data)
+    return health.run()
 
-    return app
+
+@app.route("/environment")
+@limiter.limit("1 per hour")
+def environment():
```
aaron-suarez: Would be cool to get a unit test for this too, but I'm not sure why this was necessary to be honest. I think it might have something to do with that `/metrics` endpoint I mentioned but like I said, I don't recall how all this stuff interacts with each other. Or maybe it was added as just additional information that we could possibly use in the cluster for metrics and such... Anyway, yeah if we're going to have it, might as well get a test around it.

aaron-suarez: Another thing about this.. This is out of scope for this PR, but this endpoint gives out a lot of sensitive information (luckily it masks the credentials, but it gives out all the environment variables and such). We should consider making sure this is only accessible to the cluster and not publicly.

aaron-suarez:
Once we get the unit tests and fix the `/metrics` endpoint, I'm good with merging this. Thanks again!

app/__init__.py
irvingpop: flake8 complains that all imports aren't at the top of the file, but when I moved them it broke everything because these imports need to be after the definition of `db`

app/__init__.py
irvingpop: I'm not sure how I would do that in a unit test. do you have any ideas on how to implement that?

app/__init__.py
irvingpop: then we'll need some kind of access-control plugin for Flask and the implementation of that.  How about we just comment out the `/environment` route for now since it's of no apparent value and a security risk?

app/__init__.py
aaron-suarez: Ok that's fine we can leave it in. Thanks for clarifying

app/__init__.py
aaron-suarez: Yeah we can comment it out for now

app/__init__.py
aaron-suarez: There's a rate limiting unit test already. You can follow that same pattern but change the assertion that it's still returning the expected response rather than the 429

https://github.com/OperationCode/resources_api/blob/master/tests/unit/test_routes/test_utils.py#L107

aaron-suarez: Let me know if you need additional help with this and I can write the test. Not trying to cop out and make you do it all, just have been really busy this week. I'll have some time this weekend

irvingpop: test added

app/__init__.py
irvingpop: commented out

irvingpop:
hey @aaron-suarez I'm confused - after messing around with this for a while, I realized that the /metrics endpoint _did_ work -  if you do a `make run-with-metrics` or simply spin up the environment with `docker-compose up`.   there's something about running it via `make run` that breaks the metrics endpoint and I don't know what's up with that.

I did add a test for the healthz endpoint, but can you please repro what I'm saying about the metrics endpoint?

aaron-suarez:
I don't know what happened there, perhaps I made a mistake. The `/metrics` endpoint is working for me now.

I went ahead and commented out the `/environment` endpoint for you because I think you forgot to push that change up. When CI passes I'll merge.

Thanks again for your help on this, I think it's good. Let's get this deployed ASAP so we can access the Resources API again. We'll need to follow up with a PR that pins the dependencies so we don't run into an issue like this again with a breaking change.

PR #293, opened 2020-03-22
created make commands for implementing poetry

### LINKED ISSUE #281: Implement poetry for managing dependencies
Per https://github.com/OperationCode/resources_api/pull/271#issuecomment-596150359, we will have a simpler time managing dependencies with poetry (rather than manually doing what we do now with `requirements.txt`). Poetry should be used so it's consistent with https://github.com/OperationCode/back-end

Acceptance criteria: 
1) implement poetry to manage dependencies
2) ensure that this change is essentially transparent by using the Makefile. If necessary, add a new make command, but I'd really like `make build` to continue to handle the actual installation

chynh:
## Changes
- "implement-poetry" command will install poetry and add dependencies from "requirements.txt" and also modify Dockerfile accordingly.
- "add" and "remove" command can be used to add or remove dependencies going forward.

Fixes #281


aaron-suarez:
Hey @chynh! Since this whole COVID-19 pandemic I've been preoccupied with other things and totally missed this. Thanks for picking this up, I'll review it soon!

aaron-suarez:
@chynh Hey, I took a look just now. I think there was a misunderstanding in what we wanted. We want to get a poetry.lock file actually locking down the required versions of the required packages and change the Dockerfile to install those when the image is built. Once we do that, we won't need the `requirements.txt` file anymore so you can delete that in favor of the poetry.lock file.

I like the idea of being able to add and remove packages using `make`. Can you give an example of how to use that?

chynh:
> @chynh Hey, I took a look just now. I think there was a misunderstanding in what we wanted. We want to get a poetry.lock file actually locking down the required versions of the required packages and change the Dockerfile to install those when the image is built. Once we do that, we won't need the `requirements.txt` file anymore so you can delete that in favor of the poetry.lock file.

@aaron-suarez Got it. So I believe running ``make implement-poetry`` would basically initialize poetry and install all required packages from ``requirements.txt``. After that, you would have ``poetry.lock`` and ``pyproject.toml`` in both the container and your local machine. Finally, the script will modify Dockerfile to reflect the changes, like to copy pyproject.toml and poetry.lock instead of requirement.txt, install the packages on poetry.lock file when the image is built, etc. 

I didn't yet make is so that ``requirement.txt`` would get deleted but that's something that I can include in the script.

Does this sound similar to what you had in mind?
 
> I like the idea of being able to add and remove packages using `make`. Can you give an example of how to use that?

Yeah, this is definitely not included in the original scope of the issue but while working on it, I just thought it would be useful and decided to throw it in. You could use this command to add or remove package from poetry like so: 

``make add <some package>``
``make remove <some package>``

aaron-suarez:
I guess I just don't understand the use case for why we need to run `make implement-poetry` more than once. Isn't this something that you can just run and then commit the output?

chynh:
> I guess I just don't understand the use case for why we need to run `make implement-poetry` more than once. Isn't this something that you can just run and then commit the output?

Ah okay. Yeah, that's a fair point. I guess this should really be a one-time thing rather than having a command for it. 

I can remove the ``make implement-poetry command``. So just to double-check, would you be okay with me just running this and committing the changes? I think that would roughly include modifying the Dockerfile, removing requirement.txt and adding poetry.lock file.

aaron-suarez:
Yes, go ahead and run it, remove the command from the makefile and then commit the changes. And the dockerfile changes if that doesn't happen automatically. I'll do another round of review after that to make sure it all looks good

aaron-suarez:
Oh and you can leave in the add/remove commands, that looks handy. Thanks for thinking of that

aaron-suarez:
I guess we have a conflict in requirement.txt that we'll need to fix, but it seems to run fine locally. @mc962 @irvingpop Would you guys mind taking a look at these changes as well and making sure it looks right?

Dockerfile
```diff
@@ -8,16 +8,18 @@ ENV FLASK_APP run.py
...
 RUN apt-get update \
     && apt-get install -y libpq-dev gcc \
     && rm -rf /var/lib/apt/lists/* \
-    && pip install --upgrade pip
+    && pip install --upgrade pip \
+    && pip install poetry \
+    && poetry config virtualenvs.create false
 
-RUN pip install -r requirements.txt
+RUN poetry install
```
irvingpop, line 22: One suggestion, could you please add these flags to the poetry install: `--no-dev --no-interaction --no-ansi`

irvingpop:
You'll need to git rebase your branch before this is green to merge

chynh:
I did update poetry packages as I realized I wasn't using the latest ``requirements.txt``. Let me know if there is anything else that needs to be addressed

aaron-suarez:
Ok, took another look at it and we're pretty close. Please remove the `requirements.txt` file, as we won't be needing it anymore. If we do need one for some reason, we can always generate it from the poetry.lock file using `poetry export -f requirements.txt > requirements.txt`

Once you make that change, I'll be happy to merge!

PR #299, opened 2020-04-15
created click command for checking expired url

### LINKED ISSUE #83: Check for expired URL
It would be really cool to have a way to check all resource urls programatically for a bad link.

Some sites will not allow you to do HEAD or GET requests, and in those cases we can blacklist (or whitelist the opposite).

In the future we can modify the table for the database on resources to include a whitelist or blacklist flag. 

Also in the future we can determine when this is run, but for now we can hook it into unit tests, either in a sub-module that doesn't get run all the time, or in the main tests that get run all the time.

chynh:
Fixes #83

Please let me know your thoughts as this is the first iteration.

### DIFF
--- a/app/cli.py
+++ b/app/cli.py
@@ -172,9 +172,9 @@ def register(app, db):  # pragma: no cover
             try:
                 res = requests.get(resource.url, headers=headers)
                 res.status_code > 400 and \
-                    print(f"resource_id: {resource.id} reource_url: {resource.url}")
+                    print(f"resource_id: {resource.id} | resource_url: {resource.url}")
             except Exception:
-                print(f"resource_id: {resource.id} resource_url: {resource.url}")
+                print(f"resource_id: {resource.id} | resource_url: {resource.url}")
 
         with ThreadPoolExecutor(max_workers=20) as executor:
             for resource in resources:
--- a/app/cli.py
+++ b/app/cli.py
@@ -158,7 +158,7 @@ def register(app, db):  # pragma: no cover
         pass
 
     @app.cli.command()
-    def check_bad_url():
+    def check_bad_urls():
         """check for expired resource url"""
         resources = Resource.query.all()
         headers = {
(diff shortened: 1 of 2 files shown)


aaron-suarez:
It looks pretty solid from just a glance. I'll run it soon and if it works then I'm good with it.

aaron-suarez:
I'm going to give @wimo7083 another week or so to give comments. If he's good with it or silent, then I'll merge

app/cli.py
```diff
@@ -155,6 +157,29 @@ def apikey():
         """apikey commands"""
         pass
 
+    @app.cli.command()
+    def check_bad_url():
```
wimo7083, line 161: This name can be improved. The name indicates you are only checking one url instead of a mass action.

app/cli.py
```diff
@@ -155,6 +157,29 @@ def apikey():
...
+        """check for expired resource url"""
+        resources = Resource.query.all()
+        headers = {
+            'User-Agent': 'Mozilla/5.0 (Linux; Android 6.0; Nexus 5 Build/MRA58N) \
+                           AppleWebKit/537.36 (KHTML, like Gecko) \
+                           Chrome/79.0.3945.117 Mobile Safari/537.36'
+        }
+        processes = []
+
+        def print_bad_url(resource):
```
wimo7083, line 171: When doing multithreaded work, printing your result isn't the most ideal outcome. Additionally making a new serialization format instead of using something that can be fed into the database to delete sounds like an antipattern. 

Instead of printing or not printing based on a bad result, you want return a boolean.

```python
# suggested name, pick whatever you want that doesn't involve printing. 
def filter_expired(resource):
       # check for bad url, return boolean
```

app/cli.py
```diff
@@ -155,6 +157,29 @@ def apikey():
...
+            try:
+                res = requests.get(resource.url, headers=headers)
+                res.status_code > 400 and \
+                    print(f"resource_id: {resource.id} | resource_url: {resource.url}")
+            except Exception:
```
wimo7083, line 176: Not the best idea to catch all exceptions. What if you lose internet connection? What if you write bad code? You'd only want to catch exceptions that indicate the resource is a goner.

app/cli.py
```diff
@@ -155,6 +157,29 @@ def apikey():
...
+                print(f"resource_id: {resource.id} | resource_url: {resource.url}")
+
+        with ThreadPoolExecutor(max_workers=20) as executor:
+            for resource in resources:
```
wimo7083, line 180: processes = [r for r in resources if executor.submit(filter_expired, r)]

aaron-suarez:
Alright, let's get it tightened up per Will's comments and then we'll be good to merge 😄

app/cli.py
chynh: fixed

app/cli.py
chynh: fixed

app/cli.py
chynh: fixed.

app/cli.py
chynh: fixed

aaron-suarez:
Thanks for doing this for us!

PR #303, opened 2020-05-03
Add support for `languages[]` syntax

aaron-suarez:
The front end at operationcode.org/resources uses a library which translates an array called `languages` into a query string that looks like `languages[]=somelanguage&languages[]=someotherlanguage`. This adds support for this syntax with the `[]` on the end of the query parameter name of `languages[]`.

In the docs, I did not add a separate query parameter for this, I just mentioned in the description that `languages[]` will work as well.

### DIFF
--- a/app/api/routes/resource_retrieval.py
+++ b/app/api/routes/resource_retrieval.py
@@ -38,6 +38,8 @@ def get_resources():
 
     # Fetch the filter params from the url, if they were provided.
     languages = request.args.getlist('languages')
+    if not languages:
+        languages = request.args.getlist('languages[]')
     category = request.args.get('category')
     updated_after = request.args.get('updated_after')
     paid = request.args.get('paid')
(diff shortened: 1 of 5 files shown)


kylemh:
Can I look at this on the front-end? You shouldn't have to do work like this. This is a clear client error.

I see problems from axios discussed here: https://github.com/axios/axios/issues/604

Maybe it's finally time to move to https://swr.now.sh/ (for other reasons, but it also prolly doesn't have serialization errors)

aaron-suarez:
Ok, well I'll go ahead and leave this open until we determine how difficult of a fix it is. This was super easy and I don't see any risk adding it here. But yes, I would think a front end fix makes more sense if that's similarly doable.

kylemh:
It just seems like an awful interface for you to have to prepare for. It's prolly easy for you deal with me saying `imanasshole-language,language,language` too, but that doesn't mean you should 😂

aaron-suarez:
Sounds like we got the axios issue worked out. Closing this for now. We can always re-open if necessary.

PR #306, opened 2020-05-06
Added method to ensure bool and allow strings that parse to bool

jamesfitzer:
… to bool with parsing

### DIFF
--- a/tests/unit/test_routes/test_resource_create.py
+++ b/tests/unit/test_routes/test_resource_create.py
@@ -39,6 +39,52 @@ def test_create_resource(
                                notes)
     assert_invalid_create(response, ["category", "paid", "notes"], 0)
 
+    # Boolean strings that can be parsed to true or False
+    name = "StringsForBoolsFalse"
+    url = None
+    category = None
+    languages = None
+    paid = "FaLsE"
+    notes = "Some notes"
+    response = create_resource(client,
+                               apikey,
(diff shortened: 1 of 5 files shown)


jamesfitzer:
This isn't quite ready. Need unit tests, and probably unify the update logic with the create logic (similar logic exists in update but it could use my helper function too.)

tests/unit/test_routes/test_resource_create.py
```diff
@@ -39,6 +39,52 @@ def test_create_resource(
...
+                               url,
+                               category,
+                               languages,
+                               paid,
+                               notes)
+    assert (response.status_code == 200)
+    assert response.json['data'][0].get('paid') is True
+
+    # Bad "paid" data
+    paid = "PERHAPS"
```
aaron-suarez, line 77: Thanks for adding a test here to check just the paid value is bad. This test does way too much
https://github.com/OperationCode/resources_api/blob/610fd861b320f2baa8a9f65c3986c05927119117/tests/unit/test_routes/test_resource_create.py#L25-L40

I'd like us to go back and refactor this test so that it makes a new request for every piece of invalid data, and that everything else in that request is valid. This way we know if any given invalid piece of data will cause the 422, rather than just calling it good because something in there failed validation.

Fixing that is outside of the scope of this PR though. This is good.

PR #308, opened 2020-05-09
Update the uwsgi settings with better production features

aaron-suarez:
This started due to looking at the production logs and getting overwhelmed with logs of `/healthz` over and over again. Upon further research, I found this article https://www.techatbloomberg.com/blog/configuring-uwsgi-production-deployment/

I decided to go ahead and add all of the sections I think were relevant to us right now, and we'll try it out and see what happens.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ COPY pyproject.toml poetry.lock ./
 RUN mkdir /static
 
 RUN apt-get update \
-    && apt-get install -y libpq-dev gcc \
+    && apt-get install -y libpq-dev gcc libpcre3 libpcre3-dev \
     && rm -rf /var/lib/apt/lists/* \
     && pip install --upgrade pip \
     && pip install poetry \
--- a/app.ini
+++ b/app.ini
@@ -1,5 +1,13 @@
 ; app.ini
+
+; NOTE: The configuration in this file came from the following
+; sources. These should answer the "why" of the decision for
+; each of the configuration options:
+; - https://medium.com/bitcraft/docker-composing-a-python-3-flask-app-line-by-line-93b721105777
+; - https://www.techatbloomberg.com/blog/configuring-uwsgi-production-deployment/
+
 [uwsgi]
+strict = true
 protocol = uwsgi
 ; This is the name of our Python file
 ; minus the file extension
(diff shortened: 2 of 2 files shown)


app.ini
```diff
@@ -15,3 +16,9 @@ processes = 2
 http = 0.0.0.0:5000
 vacuum = true
 die-on-term = true
+single-interpreter = true
+need-app = true
+
+disable-logging = true
```
wimo7083, line 22: I'm not sure what's best here. When I'm dealing with "too much uneeded logs", I have to think hard of removing the logs is the fix or if you want to filter logs better. Any log aggregation framework will allow you to filter logs or ingest differently. 


If you just want to disable logging for healthz, you can probably use 
`route = ^/healthz donotlog:`

https://stackoverflow.com/a/25686557

My personal preferences are to find ways to better view logs and filter `/healthz` when interacting with them. That means when you need to inspect any 2xx, 3xx, logs you'll still have them.


If this is a cost optimization then by all means reduce the logging, otherwise I'd go with my recommendation. The choice is yours though. I think the other changes are a good.


I'd also add 
```
die-on-term = true  
vacuum = true
```

app.ini
```diff
@@ -1,5 +1,6 @@
 ; app.ini
 [uwsgi]
+strict = true
```
wimo7083, line 3: Finding uwsgi configuration information is pretty hard. Please include a comment for the blog post you found these at. The PR is usually a good place for this, but the commit message or inline comments are the best.

app.ini
aaron-suarez: Lines 17 and 18 have the `die-on-term` and `vacuum` settings you recommend, yeah?

aaron-suarez: Thanks for the `donotlog` suggestion, I've been looking for that and couldn't find it. I think I'll go ahead and do that instead. To disable logging altogether isn't a huge deal though, because I explicitly log the POST and PUT requests when they come in, and I can't imagine wanting to see GET requests that are 2xx... The 3xx don't log right now anyway. Also of note, we're still on DEBUG level logging in production right now, but I'm fine keeping that until it gets unwieldy. I'll disable logging if and when we need to, but that healthz has got to go

app.ini
aaron-suarez: I will add an inline comment, thanks

app.ini
wimo7083: log level isn't really important until it is. Sounds good.

Dockerfile
```diff
@@ -13,7 +13,7 @@ COPY pyproject.toml poetry.lock ./
 RUN mkdir /static
 
 RUN apt-get update \
-    && apt-get install -y libpq-dev gcc \
+    && apt-get install -y libpq-dev gcc libpcre3 libpcre3-dev \
```
aaron-suarez, line 16: We need `pcre` support for the part of the app.ini file that says `route = ^/healthz donotlog:`

PR #325, opened 2020-06-14
Add JWT in auth header as an authN option

aaron-suarez:
Adds JWT as an authentication option (this is how the Resources page will authenticate).


aaron-suarez:
@AllenAnthes I think I got a bunch of good test cases. I filled in some areas that were lacking validation and certain error cases are handled better now. I think we're just about there. Please give it a good shake and we should be ready to merge

Makefile
```diff
@@ -74,6 +74,11 @@ test: build
 test-coverage:
 	${DOCKER_COMPOSE} run ${RESOURCES_CONTAINER} py.test --cov-report html --cov=app/ tests/
 
+# Usage: make test-single tests/unit/test_api_key.py::test_get_api_key
+.PHONY: test-single
+test-single: build
```
aaron-suarez, line 79: This has nothing to do with this PR in particular, I just realized it was useful and wanted to use it

Issue #270, opened 2020-01-10
Fix remaining issues with tests


aaron-suarez:
In #266 we made a lot of improvement on the tests, but we had to throw in the towel on a few of the issues:

1) The `search` tests might not be testing anything. See https://github.com/OperationCode/resources_api/pull/266#discussion_r361545451

2) There's almost certainly a better abstraction for `test_algolia_exception_error` and `test_algolia_unreachable_host_error` because they are the exact same test, just with different fake errors returned from algolia. Someone can go in and abstract out the similarities and perhaps inject the mocks just where they're needed to cut down on the code and/or execution time. See https://github.com/OperationCode/resources_api/pull/266#discussion_r363545482

3) We were never able to implement `test_false_validation` due to problems mocking the correct function. I'm not sure if it was our implementation or perhaps a bug in pytest-mock, but it just wasn't working for us. If you can make it work, we'd love this test to exist. See https://github.com/OperationCode/resources_api/pull/266#discussion_r363123518

aaron-suarez:
@rooshs481 is going to take this one on

fumblehool:
I think nobody is currently working on this. @aaron-suarez  Can I work on this?

aaron-suarez:
Go ahead @fumblehool !

PR #333, opened 2020-06-23
refactoring test suite

rooshs481:
refactored auth tests


Digbigpig:
Is this for issue #270 ?

This looks like a good start!

aaron-suarez:
I think it might actually be for #82, but I agree, it's a great start. I'd like to see this get over the finish line. Let me know if you need any help @rooshs481

PR #340, opened 2020-07-03
removed pagination for GET languages and GET categories

### LINKED ISSUE #328: Stop paginating on `GET /categories` and `GET /languages`
The front end needs to have all of the data available from those endpoints to populate the dropdowns. Therefore, we need to stop paginating and let the client decide how it wants to handle the data.

mike-lloyd03:
fixes #328

### DIFF
--- a/app/static/openapi.yaml
+++ b/app/static/openapi.yaml
@@ -167,7 +167,7 @@ paths:
   /categories:
     get:
       summary: List Categories
-      description: Gets a list of all available categories. Note&#58; The `page` attribute in the response of the `/search` endpoint begins with `0`, but the `page` attribute in the response of the `List Categories` endpoint begins with `1`.
+      description: Gets a list of all available categories.
       tags:
         - List Categories
       responses:
@@ -210,21 +210,16 @@ paths:
                       name: 'Docker'
                     - id: 15
                       name: 'Regular Expressions'
-                    - id: 1
+                    - id: 16
                       name: 'Mentorship'
-                    - id: 1
+                    - id: 17
                       name: 'News'
-                    - id: 1
+                    - id: 18
                       name: 'Back End Dev'
-                    - id: 1
+                    - id: 19
                       name: 'Mobile Dev'
-                    - id: 1
+                    - id: 20
                       name: 'Cheat Sheets'
-                  has_next: true
-                  has_prev: false
-                  number_of_pages: 4
-                  page: 1
-                  records_per_page: 20
                   status: 'ok'
                   status_code: 200
                   total_count: 72
@@ -265,7 +260,7 @@ paths:
   /languages:
     get:
       summary: List Languages
-      description: Gets list of all available language entries. Note&#58; The `page` attribute in the response of the `/search` endpoint begins with `0`, but the `page` attribute in the response of the `List Languages` endpoint begins with `1`.
+      description: Gets list of all available language entries.
       tags:
         - List Languages
       responses:
@@ -318,11 +313,6 @@ paths:
                     name: 'Android'
                   - id: 20
                     name: 'CSS'
-                has_next: true
-                has_prev: false
-                number_of_pages: 4
-                page: 1
-                records_per_page: 20
                 status: 'ok'
                 status_code: 200
                 total_count: 73
@@ -932,21 +922,6 @@ components:
           type: string
         category:
           $ref: '#/components/schemas/Category_data'
-        has_next:
-          type: boolean
-          description: Additional pages
-        has_previous:
-          type: boolean
-          description: Previous pages
-        number_of_pages:
-          type: integer
-          description: Total pages
-        page:
-          type: integer
-          description: Current page
-        records_per_page:
-          type: integer
-          description: Records per page
         status:
           type: string
           description: Status message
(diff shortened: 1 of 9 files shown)


aaron-suarez:
This is looking really really good! Thanks so much for putting this together!

We need some documentation updates:

This note can be taken out and the description can just be "Gets a list of all available categories."
https://github.com/OperationCode/resources_api/blob/813fccbb28421ce9ee35bb9225a3aa41b00a235f/app/static/openapi.yaml#L170

These pagination details can be taken out
https://github.com/OperationCode/resources_api/blob/813fccbb28421ce9ee35bb9225a3aa41b00a235f/app/static/openapi.yaml#L223-L227

While you're in there, idk why these IDs are all "1", they should be sequential (16, 17, 18, etc).. Can you go ahead and update those?
https://github.com/OperationCode/resources_api/blob/813fccbb28421ce9ee35bb9225a3aa41b00a235f/app/static/openapi.yaml#L213

These pagination details can be taken out too
https://github.com/OperationCode/resources_api/blob/813fccbb28421ce9ee35bb9225a3aa41b00a235f/app/static/openapi.yaml#L935-L949

Then we need similar changes to the `languages` docs in the same file. Once you make your changes to this file I'll review and let you know if we missed anything. I recommend going to localhost:5000 in your web browser and checking out the rendered docs and trying to find other references to pagination on the categories and languages endpoints. If you can figure out where those exist in the openapi file, great! If you can't, but you see something wrong in the docs, just post a screenshot and I'll direct you to the line in the Open API file.

mike-lloyd03:
I updated the docs so there's no reference to pagination in List Languages and List Categories.

app/utils.py
```diff
@@ -101,7 +101,7 @@ def standardize_response(
     """
     data = payload.get("data")
     errors = payload.get("errors")
-    pagination_details = payload.get("pagination_details")
+    pagination_details = payload.get("details")
```
aaron-suarez, line 104: This might be kind of nit-picky but now that I'm looking at it I think this is going to bother me. We should change `pagination_details` to `details` here and on lines 132 & 133, since it's not always pagination details and that could be confusing in the future.

tests/unit/test_routes/test_categories.py
```diff
@@ -10,16 +10,7 @@ def test_categories(module_client, module_db):
...
-
-
-def test_categories_page_out_of_bounds(module_client, module_db):
-    client = module_client
-    too_far = 99999999
-    response = client.get(
-        f"api/v1/categories?page_size=100&page={too_far}", follow_redirects=True)
-
-    assert_correct_response(response, 404)
+    assert (response.json['total_count'] is not None)
```
aaron-suarez, line 13: Let's also add an assertion here that we're returning all of the categories in the response (i.e., pagination is not happening). Something like this ought to work
```python
    assert (len(response.json['categories']) == response.json['total_count'])
```

app/static/openapi.yaml
```diff
@@ -318,11 +313,6 @@ paths:
...
                   - id: 20
                     name: 'CSS'
-                has_next: true
-                has_prev: false
-                number_of_pages: 4
-                page: 1
-                records_per_page: 20
                 status: 'ok'
                 status_code: 200
                 total_count: 73
```
aaron-suarez, line 318: This total count should not be 73 unless we add 73 languages to the above list.. I'm thinking we make it match what's in the local development DB right now, which is 23, and make the list match the 23 langauges we have in there.

app/static/openapi.yaml
```diff
@@ -210,21 +210,16 @@ paths:
...
+                    - id: 20
                       name: 'Cheat Sheets'
-                  has_next: true
-                  has_prev: false
-                  number_of_pages: 4
-                  page: 1
-                  records_per_page: 20
                   status: 'ok'
                   status_code: 200
                   total_count: 72
```
aaron-suarez, line 225: Similar issue here, let's make this 23 and add the 23 categories that are in the local db right now to the list above

aaron-suarez:
I also noticed while I was looking at the docs that this line is incorrect
https://github.com/OperationCode/resources_api/blob/813fccbb28421ce9ee35bb9225a3aa41b00a235f/app/static/openapi.yaml#L977

Can we make this `languages` like it should be, instead of `language` while we're in here? Thanks!!

aaron-suarez:
I think once those changes are made, we're good to go 👍  Thanks again for your help on this 😄

mike-lloyd03:
Thanks for the detailed corrections. I'll try to get this buttoned up by the end of the day on Monday.

app/static/openapi.yaml
mike-lloyd03: I just noticed we have Linux listed twice. Should I include it anyway or leave it off the list and make the total count 22?

```json
{
  "id": 3, 
  "name": "Unix/Linux/*nix"
}
```

```json
{
  "id": 23, 
  "name": "Linux"
}
```

aaron-suarez: It's because we have it twice in the resources.yml

https://github.com/OperationCode/resources_api/blob/main/resources.yml#L87

https://github.com/OperationCode/resources_api/blob/main/resources.yml#L1263

Would be useful to change that if we change this. So it's up to how motivated you are to change it all I guess

mike-lloyd03: It seems appropriate to change all the `"Linux"` resources to `"Unix/Linux/*nix"`. Changing _The Art of Unix Programming_ to `"Linux"` wouldn't really be accurate. I can add this to my next commit.

aaron-suarez: Do you want to change it in prod too? That would be helpful I would think.

mike-lloyd03:
Okay forgot to run the tests before committing and had a copy-paste mistake. Fixed it now. I think we've got every box ticked.

PR #341, opened 2020-07-03
fixes #338

### LINKED ISSUE #338: Return "Getting Started" resources first when no parameters specified
When the user first arrives at the Resources page on the front end, they are presented with whatever resources the database happens to serve up. Since we don't yet have featured categories on the front end, it would be nice to go ahead and serve up some resources that are relevant for beginners who maybe have no idea where to start on their learning journey. There are a bunch of great blog posts and similar "Getting Started" materials in the `Getting Started` category. From there, they can search for what they want, or filter by categories/languages. 

Acceptance criteria: When a request is made without any filter parameters to the `GET /resources` endpoint, the API should return "Getting Started" resources, followed by all the other resources available (paginated).

This should be able to be accomplished by adding an ORDER BY clause in this particular case. I have an idea adapted from this [stackoverflow post](https://stackoverflow.com/a/6332081):
```SQL
ORDER BY
   CASE category
      WHEN 'getting started' THEN 1
      ELSE 2
   END, id
```

I have no idea how to do this with Flask-SQLAlchemy, but we'll figure it out. Importantly, we shouldn't add this ORDER BY clause unless there are no other parameters at all. So we'll want some unit tests around this. I don't think it needs to be added to the documentation so we can skip making changes to the openapi.yaml file.

jtmst:
fixes #338

### DIFF
--- a/app/api/routes/resource_retrieval.py
+++ b/app/api/routes/resource_retrieval.py
@@ -2,7 +2,7 @@ from datetime import datetime
 
 from dateutil import parser
 from flask import redirect, request
-from sqlalchemy import func, or_
+from sqlalchemy import func, or_, text
 
 from app import utils as utils
 from app.api import bp
@@ -44,6 +44,7 @@ def get_resources():
 
     q = Resource.query
 
+
     # Filter on languages
     if languages:
         # Take the list of languages they pass in, join them all with OR
@@ -60,6 +61,7 @@ def get_resources():
                 func.lower(Category.name) == category.lower()
             )
         )
+
 
     # Filter on updated_after
     if updated_after:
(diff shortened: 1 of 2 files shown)


app/api/routes/resource_retrieval.py
```diff
@@ -44,6 +44,7 @@ def get_resources():
 
     q = Resource.query
 
+
```
aaron-suarez, line 47: I think this extra line and the other one on line 65 are causing the linter to fail. You can run `make lint` locally to check it out and then fix whatever errors it tells you about

aaron-suarez:
Oh, you know what? The CI failed on the tests, not the linting (though I still think we should remove the extra lines).

The test that failed checks to make sure that resource ID 1 is the first one returned, which is no longer expected behavior. We should change that test to check for the category to equal "Getting Started" rather than checking the ID. It's probably good to get a unit test around this anyway so that if someone changes the behavior we know about it and it's intentional.

If you want to screen share on it to fix the issues we can, just hit me up on slack.

jtmst:
That sounds good to me but I likely won't be around long enough to do it tomorrow.  Ill be around Sunday/Monday though. I'll ping you in slack sometime then.

Thanks!

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

aaron-suarez:
Enjoy your 4th! I will also likely not be around due to festivities, but we'll sync up soon

PR #342, opened 2020-07-04
Change blacklisted -> denied for inclusivity

aaron-suarez:
The word `blacklist` isn't a very inclusive term. I think that `denylist` is a much better term to use, and besides, it's more descriptive for what we're doing to the API key.

This PR involves a DB migration and also has everything to do with our authentication, so please give it a solid review. To run the migration locally, you should be able to run `make migrate`. Let me know if this doesn't work, because we'll want to fix that command. And of course, let me know if you run into any issues.

Testing:
1) Get your API key with the `/apikey` endpoint
2) Use the API key to successfully add a resource with `POST /resources`
3) Use the API key to successfully update a resource with `PUT /resources`
4) Deny your API key using `docker-compose run resources-api flask apikey deny <email address>` (can also pass in API key instead of email address)
5) Ensure you can't get your API key from the `/apikey` endpoint after it's been put on the denylist
6) Ensure you can't make a `POST` or `PUT` request to the `/resources` endpoint with your denied API key
7) Reactivate your API key using `docker-compose run resources-api flask apikey reactivate <email address>` (can also pass in API key instead of email address)
8) Get your API key again and make sure you're allowed to make POST requests again.


aaron-suarez:
@jamesfitzer I can't add you as a reviewer until you accept the invite to collaborate, but would you mind also taking a look when you get a chance?

jamesfitzer:
I’ve accepted twice now . Something isn’t working

Will look at this Monday

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

jamesfitzer:
Ok should be good now

migrations/versions/205742d3b3f5_change_blacklist_to_deny.py
```diff
@@ -0,0 +1,31 @@
...
+revision = '205742d3b3f5'
+down_revision = '4b6587b98b26'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('key', sa.Column('denied', sa.Boolean(), nullable=True))
+    op.drop_column('key', 'blacklisted')
```
toadzky, line 23: you should probably add a migration to populate the new column with the old values. same with the downgrade

migrations/versions/205742d3b3f5_change_blacklist_to_deny.py
```diff
@@ -0,0 +1,29 @@
...
+# revision identifiers, used by Alembic.
+revision = '205742d3b3f5'
+down_revision = '4b6587b98b26'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.alter_column('key', 'blacklisted', new_column_name='denied')
```
aaron-suarez, line 22: I tested out upgrading and downgrading with a denied key, and it kept the attribute both ways with this, so I think we're good on that now.

aaron-suarez:
Reviews were dismissed because I had to rebase to get back up to date with `main` but the content of the PR is the same. Just need an approval and merge 👍

PR #343, opened 2020-07-06
Raise DBAPIError rather than None

aaron-suarez:
When we merged #341, the tests started failing on this line
https://github.com/OperationCode/resources_api/blob/b4d281c2d8215162309a2db3f66d20d13fe8716b/app/api/routes/resource_retrieval.py#L91

I'm not sure exactly why this fixes it, but it's more correct anyway and it happens to fix the tests. I'm going to merge it and then do some manual testing on staging and make sure everything is good to go before we deploy to prod.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,6 +4,7 @@ from algoliasearch.exceptions import (AlgoliaException,
 from app import app
 from app import db as _db
 from app.utils import standardize_response
+from sqlalchemy.exc import DBAPIError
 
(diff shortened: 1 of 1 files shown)


aaron-suarez:
I did some manual testing in staging and everything looks good to me. These changes didn't affect anything with auth or with creating/updating resources. All the GET requests return the data I'm expecting. So I really think it was just a weird test thing and not anything to worry about in production. I'm going to go ahead and deploy 🤞

PR #348, opened 2020-07-10
CircleCI: automatically deploy main branch to staging on every merge

aaron-suarez:
When a new PR is merged into `main`, I would like to merge that to the staging environment automatically, so we can test right away.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,18 @@
 version: 2.1
+parameters:
+  aws-profile:
+    type: string
+    default: default
+  aws-region:
+    type: string
+    default: us-east-2
+  cluster-name:
+    type: string
+    default: operationcode-backend
 
 orbs:
   docker: circleci/docker@0.5.1
+  aws-eks: circleci/aws-eks@0.2.7
 
 workflows:
   build_test_deploy:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -26,6 +27,18 @@ workflows:
               only:
                 - main
 
+      - aws-eks/update-container-image:
```
irvingpop, line 30: This should be in the steps section below

.circleci/config.yml
```diff
@@ -64,4 +77,12 @@ jobs:
       - run:
           name: Run Bandit security analysis
           command: |
-            make bandit
\ No newline at end of file
+            make bandit
+
+  ????:
```
irvingpop, line 82: This is the job name. Maybe call it staging_deploy?

.circleci/config.yml
```diff
@@ -64,4 +65,29 @@ jobs:
...
+          Operation Code Backend Cluster
+        type: string
+    steps:
+      - update-kubeconfig-with-authenticator:
+          aws-profile: << parameters.aws-profile >>
+          aws-region: << parameters.aws-region >>
+          cluster-name: << parameters.cluster-name >>
+          install-kubectl: true
+      - aws-eks/update-container-image:
+          cluster-name: operationcode-backend
```
irvingpop, line 84: I don't think this needs to be here

.circleci/config.yml
```diff
@@ -64,4 +75,28 @@ jobs:
       - run:
           name: Run Bandit security analysis
           command: |
-            make bandit
\ No newline at end of file
+            make bandit
+
+  staging_deploy:
+    executor: aws-eks/python3
```
irvingpop, line 81: This should be just python3

.circleci/config.yml
```diff
@@ -1,7 +1,18 @@
...
+parameters:
+  aws-profile:
+    type: string
+    default: default
+  aws-region:
+    type: string
+    default: us-east-2
+  cluster-name:
+    type: string
+    default: operationcode-backend
```
irvingpop, line 11: I think all of these parameters are going to need to move down inside the job (currently line 82)

PR #349, opened 2020-07-10
Use CIRCLE_SHA1 as the image tag instead of BUILD_NUM which is unreliable

irvingpop:
The problem here: https://app.circleci.com/pipelines/github/OperationCode/resources_api/524/workflows/39d5b471-abef-4067-b6a1-50d021043875/jobs/648

was that using CIRCLE_BUILD_NUM is unreliable because each workflow job will have a new number.   This PR proposes using the git commit sha as the tag instead, which is more reliable and traceable back to git!

Signed-off-by: Irving Popovetsky <irving@popovetsky.com>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,11 +15,11 @@ workflows:
           tag: latest
           after_build:
             - run:
-                name: Tag and Push with $CIRCLE_BUILD_NUM
+                name: Tag and Push with $CIRCLE_SHA1
                 command: |
                   IMAGE_ID=operationcode/resources-api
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -15,11 +15,11 @@ workflows:
           tag: latest
           after_build:
             - run:
-                name: Tag and Push with $CIRCLE_BUILD_NUM
+                name: Tag and Push with $CIRCLE_SHA1
```
aaron-suarez, line 18: Yeah, Ive always thought it was weird that we used the build number instead of the git SHA. I'm good with this change 👍

PR #362, opened 2020-08-16
Add check to resource_modification for orphaned langs/cats

### LINKED ISSUE #344: Languages/Categories should delete themselves if they no longer exist
Suppose we have a language called `CSS/Bootstrap` and then someone realized that that's two languages and then changes the languages for that resource to be `["CSS","Bootstrap"]` instead. What happens to the language `CSS/Bootstrap`? Well, it gets orphaned. It exists, but there are not resources that have that language.

I think that whenever a resource is updated and the `languages` attribute is one of the changes, we should check that the old languages still exist on some resource. If they do not, then we should delete the language.

Same thing with categories. If a category gets removed from the last resource with that category, then the category should be deleted as well.

mike-lloyd03:
I'm unable to test anything without an Angolia key but I put this together just to see if I'm on the right track.

Fixes #344


aaron-suarez:
You can actually sign up for a free trial of Algolia here https://www.algolia.com/users/sign_up

That will give you an API key that you can use for testing this kind of thing if you want. But we should be able to make sure most of it works how we need it to without the algolia part, and then we can test out the algolia stuff in staging and if there's a problem just fix it before we go to prod

app/api/routes/resource_modification.py
```diff
@@ -46,11 +46,19 @@ def update_resource(id, json, db):
         logger.info(
             f"Updating resource. Old data: {json_module.dumps(resource.serialize)}")
         if json.get('languages'):
+            old_languages = resource.languages[:]
             resource.languages = langs
             index_object['languages'] = resource.serialize['languages']
+            for language in old_languages:
+                if not Resource.query.filter_by(languages=language).all():
```
aaron-suarez, line 53: I don't like a SQL SELECT being inside a for loop at all. We should be able to select everything we need just once and work with that, right?

It would require more memory to store all the Resources in an array and then search for the specific ones we need.. not sure if the trade-off of time for more memory is the right answer either... 

I think optimizing this will require more research and thinking it through. And perhaps it's not a huge deal because this route isn't used very often, maybe we can worry about optimizing it later.

If we do decide to kick that can down the road, we should leave a comment about it's possible performance issues so when someone (us?) is trying to hunt down the cause of a performance problem later, it's easier to find.

mike-lloyd03: I'd rather get it right now. I can change this to query the database once.

aaron-suarez:
I took a look at the failing tests. I have a suspicion about the problem with a couple of them, they look legit. If you want to jump on a call to troubleshoot those, just let me know

mike-lloyd03:
@aaron-suarez I am trying another approach at this PR which doesn't have a query in a for loop. I've only implemented the check on the Category modification just to see what you think so far. If you think it looks good. I'll go ahead and apply the logic to the Language modification.

app/api/routes/resource_modification.py
```diff
@@ -49,8 +53,19 @@ def update_resource(id, json, db):
             resource.languages = langs
             index_object['languages'] = resource.serialize['languages']
         if json.get('category'):
+            old_category = resource.category #I'm assuming this is the category name as a string
```
aaron-suarez, line 56: This is not the category name as a string. This is a `Category` object. This is the definition of a `Category` object:
https://github.com/OperationCode/resources_api/blob/a78426a89e4625a733763c4dc24c978d639f8c52/app/models.py#L106-L130 

To get the string representation of the category, you'll want `resource.category.name`

app/api/routes/resource_modification.py
```diff
@@ -49,8 +53,19 @@ def update_resource(id, json, db):
...
+            resource_categories = get_all_resource_categories_as_strings()
             resource.category = categ
             index_object['category'] = categ.name
+            # Check if the old category is being used anywhere else
+            # in the Resources table
+            orphaned_category = Category.query \
+                .filter(Category.name.notin_(resource_categories)) \
+                .filter(Category.name == old_category) \
+                .first()
```
aaron-suarez, line 65: Does this actually need to be a SQL query? Can we not just loop through the array of strings `resource_categories` and look for `old_category` and if it's not in there, we know that `old_category` is orphaned?

I haven't thought through it a whole lot, so if the answer is we do indeed need a SQL query then I'm willing to hear about why. But at first glance it looks like we don't

mike-lloyd03: Since `old_category` is a Category object, we can just call `old_category.delete()` on it if it's orphaned right? In that case, we wouldn't need the query you highlighted above. I can just tweak the following `if` statement and be done with it. I'll have another commit tonight.

aaron-suarez: I think you delete with `db.session.delete(old_category)`

mike-lloyd03: Thanks for saying "I think" so I don't look as dumb for saying that.

mike-lloyd03:
Okay I think we're good to go now. I cleaned up the logic and implemented similar logic for the Languages portion.

aaron-suarez:
Can we get some unit tests around it? We want to make sure that an unused Category is deleted, an unused language is deleted, and that no categories or languages that are still in use are deleted

mike-lloyd03:
@aaron-suarez Okay I cherry-picked from @king-11 's tests branch but CircleCI is still failing. I'll take a look later tonight.

tests/unit/test_routes/test_resource_update.py
```diff
@@ -256,29 +254,56 @@ def test_update_resource(
     assert_correct_response(response, 404)
 
 
-def test_update_votes_authorization_header(
-        module_client, module_db, fake_auth_from_oc, fake_algolia_save):
```
aaron-suarez, line 260: I'm going to need this test to be added back in. I'm pretty sure it was deleted due to the cherry-picking situation, not an intentional delete. In any case, please just add it back in somewhere in the file because this is a very important test

mike-lloyd03:
Okay I fixed the botched cherry-pick with more commits than necessary. But hopefully this is the end.

tests/unit/test_routes/test_resource_update.py
```diff
@@ -282,3 +282,58 @@ def test_update_votes_authorization_header(
     assert (response.status_code == 200)
     assert (response.json['resource'].get(f"{UPVOTE}s") == initial_upvotes)
     assert (response.json['resource'].get(f"{DOWNVOTE}s") == initial_downvotes + 1)
+
+
+def test_delete_unused_languages(module_client, module_db,
```
aaron-suarez, line 287: Can we get a test around deleting unused category as well? Would look just like this but be `test_delete_unused_category` and then of course it would test that the unused category was deleted.

After that, this is ready to merge!

aaron-suarez (approved):
Love it! Thanks for your help, y'all!

PR #382, opened 2020-10-01
Renamed all references of paid to true

shivamk01:
Cannot delete last language when updating resource


Kandeel4411:
I think this is spam/should be reported @aaron-suarez . He seems to copy random descriptions from commits/issues and merge other branches, probably just for hacktoberfest count. His history seems to indicate that he has been doing that in the previous hacktoberfests as well

aaron-suarez:
Thanks for reporting it @Kandeel4411

PR #391, opened 2020-10-06
Remove log directory creation

platipo:
I didn't open a new issue because it didn't seem necessary, but #388 is blocked by the log directory creation.
I've gone though the code, checked where it was created and removed the related lines.


aaron-suarez (approved):
LGTM, thanks!

PR #392, opened 2020-10-06
Remove rate limiting for now

aaron-suarez:
It would be nice to implement rate limiting at the kubernetes level because the app doesn't know what the actual IP address of the client is. There's probably a way to use proxy_pass to make this happen, but we're not really sure what actual rate limits would indicate malicious traffic.

For now, rate limiting seems to be causing more problems than it solves. We're going to take it out and see about other solutions down the line


irvingpop (approved):
Agreed

kylemh (approved):

aaron-suarez:
Looks like a quorum to me. I'm going to go ahead and merge

PR #394, opened 2020-10-09
Add test for authorization header with upvote/downvote

aaron-suarez:
We discovered an auth error in https://github.com/OperationCode/front-end/pull/1283 where the API doesn't know how to handle an upvote or downvote without an API key sent in the header. This should fix that issue.


aaron-suarez:
Closed in favor of #395

PR #396, opened 2020-10-11
Add uid and gid as Docker environment variables

platipo:
Reference: https://github.com/OperationCode/resources_api/pull/388#issuecomment-706335715

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,8 @@ ENV PYTHONUNBUFFERED 1
 ENV PIP_NO_BINARY psycopg2
 ENV FLASK_SKIP_DOTENV 1
 ENV FLASK_APP run.py
+ENV UID 1000
+ENV GID 1000
 
 WORKDIR /src
 
@@ -23,7 +25,7 @@ RUN poetry install --no-dev --no-interaction --no-ansi
 
 COPY . /src
 
-RUN useradd --no-create-home --system -s /bin/false --uid 5000 uwsgi
+RUN useradd --no-create-home --system -s /bin/false --uid $UID --gid $GID uwsgi
 
 RUN chown -R uwsgi /src
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,6 +3,7 @@ version: 2.1
 orbs:
   docker: circleci/docker@0.5.1
   aws-eks: circleci/aws-eks@0.2.7
+  dotenv: anilanar/dotenv@1.0.0
 
 workflows:
   build_test_deploy:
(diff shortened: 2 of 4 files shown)


platipo:
Hi @aaron-suarez, I added the dotenv orb an it seems working. Do you have any feedback?

aaron-suarez:
Why is it 1000 sometimes and 5000 other times? Its not a problem I just don't understand what difference it makes

platipo:
I wasn't sure if all was correct so I chose to set the environment default values to a different number if it had raised an error I would have confirmed that the variables weren't propagating correctly.

platipo:
I can just set them all to 5000 if you want. Also do you want me to keep also the `gid`?

aaron-suarez:
It's fine how it is if you just want to leave it. If you feel like removing the gid, you can as long as it still works

platipo:
I'm ok with all the changes, do you prefer that Dockerfile default variables are 5000?

aaron-suarez:
It's fine how it is. I'm mobile right now but when I get the chance to run it and make sure everything works, I'll approve and merge if nothing is broken. Thanks for this!

platipo:
Hi @aaron-suarez, did you manage to try the changes?

Dockerfile
```diff
@@ -23,7 +25,8 @@ RUN poetry install --no-dev --no-interaction --no-ansi
 
 COPY . /src
 
-RUN useradd --no-create-home --system -s /bin/false --uid 5000 uwsgi
+RUN groupadd --gid $GID uwsgi \
+    && useradd --no-create-home --system -s /bin/false --uid $UID --gid $GID uwsgi
```
aaron-suarez, line 29: When you exec into the container, it's a little weird not having a name. Maybe we should give it one?

```
$ docker exec -it a536b7b91577 /bin/bash
I have no name!@a536b7b91577:/src$ whoami
whoami: cannot find name for user ID 5000
I have no name!@a536b7b91577:/src$
```

platipo: I'm very consufed by the missing name because both the `uwsgi` is specified as group and user name and in the container I just built the container and this is the output:
```
platipo:~/resources_api $ sudo docker run -p80:5000 -it --rm resources_api/uwsgi /bin/bash
uwsgi@e26da29e625a:/src$ whoami
uwsgi
```

aaron-suarez: I ran my container with `make run`, I wonder if that has something to do with it?

aaron-suarez: It does `docker-compose run -p 5000:5000 resources-api flask run -h 0.0.0.0`. If you do that, do you also find that you're missing the user?

https://github.com/OperationCode/resources_api/blob/34e7f7033cc6dbf559f0c057c16c51765c2cdb32/Makefile#L58

platipo: I tried some solutions, but I have no clue how to solve this :cry:

aaron-suarez: Ok when I have some time I'll see if I can solve it. Maybe it's worth merging and opening that as a separate issue... I'll want to give it one more good review before I determine that. Thanks for your patience on this

PR #400, opened 2020-10-18
Open api Structure Names

### LINKED ISSUE #259: OpenAPI file naming convention issues
We use an OpenAPI file to document our API, and there are a few issues with the naming conventions. As a matter of fact, it's almost like there are no naming conventions at all. Camel case and snake case are combined together in an odd fashion, and some of the names aren't very descriptive. Now that the document is formed, it would be good to take a look at the file as a whole and fix the naming issues.

See https://github.com/OperationCode/resources_api/pull/215#discussion_r352617465 and https://github.com/OperationCode/resources_api/pull/215#discussion_r352618239 for examples of the issues.

king-11:
- Change snake case names to PascalCase
- Change a few variable names to be more clear on what they are

closes #259

### DIFF
--- a/app/static/openapi.yaml
+++ b/app/static/openapi.yaml
@@ -253,7 +253,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/CategoryData'
+                $ref: '#/components/schemas/CategoryDataList'
               example:
                 apiVersion: '1.0'
                 category:
@@ -733,7 +733,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/ErrorList'
+                $ref: '#/components/schemas/ErrorResponse'
               example:
                   apiVersion: '1.0'
                   errors:
@@ -830,14 +830,14 @@ components:
             message:
               description: Error message
               type: string
-    ErrorList:
+    ErrorResponse:
       type: object
       properties:
         apiVersion:
           type: string
           description: API version
         errors:
-          $ref: '#/components/schemas/ErrorListData'
+          $ref: '#/components/schemas/ErrorDataList'
         status:
           type: string
           description: Status message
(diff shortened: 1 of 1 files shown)


king-11:
Formatting was enabled which did a bit of issue so had to push again

app/static/openapi.yaml
```diff
@@ -253,7 +253,7 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Category_data'
+                $ref: '#/components/schemas/CategoryDataList'
```
aaron-suarez, line 256: `GET /category/<id>` doesn't actually return a list, so I'm not sure why it would reference a schema called `CategoryDataList` now... I'm reluctant to continue reviewing this PR because the first one appears to make it worse rather than make it better.

Can you please post some screenshots of the before and after so we can see how this changes the way the docs are rendered at `http://localhost:5000`?

king-11: okay sure

king-11: @aaron-suarez I did check again and it seems there isn't any schema that matches /categories/{id}

king-11:
Screenshot_20201020_112636
Screenshot_20201020_112744
Screenshot_20201020_112801
Screenshot_20201020_112813
Screenshot_20201020_112830
Screenshot_20201020_113548
Screenshot_20201020_113830
Screenshot_20201020_113841
Screenshot_20201020_113857

king-11:
Update for Categories using 75a0720

Screenshot_20201020_115244
Screenshot_20201020_115255

king-11:
@aaron-suarez any review on this :sweat_smile:

aaron-suarez:
@king-11 I haven't forgotten about this but I haven't had time to give it the attention it deserves. As you're aware, the docs are probably the most important part because the API is unusable without good docs. I'll look and have some comments soon

aaron-suarez (approved):
Thanks for this @king-11

PR #404, opened 2020-10-20
test for checking deletion of orphaned langs

### LINKED ISSUE #344: Languages/Categories should delete themselves if they no longer exist

king-11:
random name for language use to prevent collision
test for #362 to close #344

### DIFF
--- a/tests/unit/test_routes/test_resource_update.py
+++ b/tests/unit/test_routes/test_resource_update.py
@@ -282,3 +282,57 @@ def test_update_votes_authorization_header(
     assert (response.status_code == 200)
     assert (response.json['resource'].get(f"{UPVOTE}s") == initial_upvotes)
     assert (response.json['resource'].get(f"{DOWNVOTE}s") == initial_downvotes + 1)
+
+
+def test_delete_unused_languages(module_client, module_db, fake_auth_from_oc, fake_algolia_save):
+    client = module_client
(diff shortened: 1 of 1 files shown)


tests/unit/test_routes/test_resource_update.py
```diff
@@ -1,3 +1,4 @@
+from tests.unit.test_response_util import client
```
aaron-suarez, line 1: Is this line necessary? I think later on in the file `client` is going to refer to `module_client` which is passed in by pytest, so this is actually an unused import, right?

king-11: yeah it is indeed unused

king-11 (commented):
made the requested changes

aaron-suarez:
Closed via #362

PR #408, opened 2020-10-30
Include vote direction

chynh:
## Description

Modify resource retrieval logic so that the payload includes the current vote direction on a resource for logged in users. This would allow the front-end to display the current vote direction for a user. 

I noticed that this piece of information is missing while working on a front-end issue so I jumped on it. Feel free to provide suggestions/comments.


app/api/routes/resource_retrieval.py
```diff
@@ -42,7 +43,19 @@ def get_resources():
     updated_after = request.args.get('updated_after')
     free = request.args.get('free')
 
+    apikey = request.headers.get('x-apikey')
```
aaron-suarez, line 46: I think this needs to be `g.auth_key` to work with users logged in from the front end. I ran into this problem #395

chynh: I think in #395 it made sense to do `g.auth_key` since it was already set during `@authenticate` decorator. This endpoint doesn't have that decorator so `g.auth_key` would always result in a key exception since it was never defined.

Can you double-check what your set up was? like did you get the authentication token from the backend running on localhost and used that token to hit resources API also running on localhost?

aaron-suarez: I'll admit I haven't ran it yet, just gave a quick look over the code and it jumped out at me. I'll let you know when I try it if it works, I just don't know how it would if you're only looking for the `x-apikey` header and not the authorization token

chynh: I see. That is taken care of on line 48 where it will first attempt to get the key via x-apikey header or via authentication header. That's literally copy and paste from authenticate decorator.

I couldn't test myself cuz I don't have backend setup locally. Let me know if you run into any problem.

aaron-suarez: Ok well sorry I didn't run through it before commenting 😅 I'll try to test it locally today if I get the chance but it's Halloween so I'm not sure if I will. In the next couple days for sure

app/api/routes/resource_retrieval.py
```diff
@@ -42,7 +43,19 @@ def get_resources():
...
+    filters = {'apikey': apikey, 'denied': False}
+    key = Key.query.filter_by(**filters).first() if apikey else jwt_to_key()
```
aaron-suarez, line 48: Ok yeah, this is the part that I overlooked and missed earlier. The `jwt_to_key()` handles a user logged in via the cookie from the backend. I'll resolve this and the other comment. Sorry about that

aaron-suarez (commented):
Ok, it's starting to look really good. For one thing, we need to include the `user_vote_direction` when we get a single resource as well as multiple resources. I know that the front end doesn't currently use this functionality, but if it ever does, I don't want to have to come and fix it. Why don't we add `user_votes` as part of the serialization in the Resources model, and if there is no auth key given, we just serialize to `None`?

This would also return the `user_vote_direction` when you create or update a resource, and when you retrieve a resource without having to have that logic in multiple files.

I'd also like to see some unit tests that verify `user_vote_direction` is null when there's no auth key, that it's `"None"` where there is an auth key but no vote direction, and that it's `"upvotes"` when it's been upvoted, and `"downvotes"` when it's been downvoted.

And since this is a change to the API itself, we'll need to add it to the documentation (OpenAPI file).

All that being said, I like the idea and I like where you're going with it. Let's just get it in the right place and get tests and docs for it, and it should be good to merge if there are no bugs that I can find. Thanks for this!

app/api/auth.py
```diff
@@ -146,6 +146,21 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def set_api_key(func):
```
aaron-suarez, line 149: This function seems to replicate a lot of the logic from above. Is there any way we can make this a little more DRY?

app/static/openapi.yaml
```diff
@@ -398,6 +398,7 @@ paths:
                   times_clicked: 0
                   upvotes: 0
                   url: 'http://www.freetechbooks.com/'
+                  user_vote_direction: null
```
aaron-suarez, line 401: Thanks for adding the `user_vote_direction` in the examples. We'll also need to add it (and a description) to the ResourceData schema
https://github.com/OperationCode/resources_api/blob/b2149912b71c6fe04daa4901d75c2a825899c0f6/app/static/openapi.yaml#L1061-L1100

tests/unit/test_routes/test_resource_update.py
```diff
@@ -23,6 +24,7 @@ def test_update_votes(module_client, module_db, fake_auth_from_oc, fake_algolia_
 
     assert (response.status_code == 200)
     assert (response.json['resource'].get(f"{UPVOTE}s") == initial_upvotes + 1)
+    assert (response.json['resource'].get(USER_VOTE_DIRECTION) == UPVOTE)
```
aaron-suarez, line 27: Tests look good, thanks!

app/api/auth.py
chynh: modified the `authenticate` decorator for resource retrieval and removed `set_api_key`

chynh:
Not sure if you want to reset db since this involves schema change in vote_information.

app/models.py
```diff
@@ -200,6 +209,6 @@ def __repr__(self):
 class VoteInformation(db.Model):
     voter_apikey = db.Column(db.String, db.ForeignKey('key.apikey'), primary_key=True)
     resource_id = db.Column(db.Integer, db.ForeignKey('resource.id'), primary_key=True)
-    current_direction = db.Column(db.String, nullable=False)
+    current_direction = db.Column(db.String, nullable=True)
```
aaron-suarez, line 212: I think this is going to require a db migration, do you know how to generate one?

app/models.py
```diff
@@ -63,6 +63,15 @@ def serialize(self):
             'last_updated': updated
         }
 
+    def serialize_with_vote_direction(self, apikey=None):
```
aaron-suarez, line 66: Since we're adding it to the docs that all Resources now have a vote direction, we should just add this logic into the `serialize` function rather than having a separate method for it. Doing it this way doesn't seem to have much benefit, but it does add in the problem of confusion of which method to use when, and also a response that should have the vote direction might just call `serialize` instead on accident. Example: https://github.com/OperationCode/resources_api/blob/d40d585409116f4e74c92c1c062149182a338bf0/app/api/routes/resource_modification.py#L190

app/models.py
chynh: Does running `make migrate` take care of this? Does that remove all the existing vote information? Asking since due to a change in `update_votes`, there could some unexpected bugs

app/models.py
chynh: Okay. This involves touching more files but I guess I can just have `serialize`

app/api/routes/resource_modification.py
```diff
@@ -99,7 +101,9 @@ def get_unique_resource_languages_as_strings():
         db.session.commit()
 
         return utils.standardize_response(
-            payload=dict(data=resource.serialize),
+            payload=dict(
+                data=resource.serialize(g.auth_key.apikey)
```
aaron-suarez, line 105: Nit: isn't this in the same function as above, and so we can reference it as `apikey` rather than `g.auth_key.apikey`?

app/models.py
aaron-suarez: No, `make migrate` runs a migration that's been created but hasn't been applied to your local db. It's what developers will have to do for their local environments after we merge. 

Though I will say, we should probably rename that as `make upgrade` and get rid of the migrate command in there because I don't think it does what I thought it did when I wrote that into the Dockerfile.

You'll want to name your migration, so you'll do something like `docker-compose run resources-api flask db migrate -m "include vote direction"` and then you'll need to check it to ensure that everything worked correctly in the auto generated file, then upgrade your database with `docker-compose run resources-api flask db upgrade` and if it successfully upgrades the database, then you can commit that file.

See also https://flask-migrate.readthedocs.io/en/latest/#example

chynh: I ran the migration command but saw the message below. 

Screen Shot 2020-11-05 at 9 21 49 PM

aaron-suarez: Hmm, I saw something different when I ran it. Maybe I'll just go ahead and generate the file and push it up to your branch

aaron-suarez:
I notice that the `/click` endpoint isn't returning the vote direction correctly. I haven't dug any deeper, and I know that endpoint is not really in use on the front end, but I'd like to see that working properly before we merge because it could cause other bugs for other endpoints depending on what the root cause of it is

chynh:
> I notice that the `/click` endpoint isn't returning the vote direction correctly. I haven't dug any deeper, and I know that endpoint is not really in use on the front end, but I'd like to see that working properly before we merge because it could cause other bugs for other endpoints depending on what the root cause of it is

If it is the issue where vote direction is always returning None, that is expected as that is not an autheticated route. If you want the vote direction when logged in, I think we can have the decorator on that route similar to what we have on retrieval routes.

aaron-suarez:
Ah, yes I see now. Can we go ahead and add that decorator to all routes that have a resource in the response so the API response is the same across the board?

aaron-suarez (approved):

PR #415, opened 2020-12-16
Remove Hacktoberfest banner

wendeee:
I thought it would be nice to remove the Hacktoberfest banner image since the event is over:)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,15 +1,12 @@
 <div align="center">
   <a href="https://operationcode.org" height="750" width="750">
-    <img
-      alt="Operation Code Hacktoberfest Banner"
-      src="https://operation-code-assets.s3.us-east-2.amazonaws.com/operationcode_hacktoberfest_2020.jpg"
-    >
+   
   </a>
 </div>
 <br />
 <br />
 
-# 🎃 Hacktoberfest 🎃
+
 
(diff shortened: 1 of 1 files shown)


C-Lion:
I think since members participated instead of deleting we should run a series of short articles highlighting those who is did (why, how they got involved, etc) and linking to their GH and LI accounts. These articles can then be used as Tweets and Linkedin posts with links back to Operation Code's and our mission to help vets learn to code. Valid Hacktoberfest PRs are proof positive of this ability.

README.md
```diff
@@ -1,15 +1,12 @@
...
+   
   </a>
 </div>
 <br />
 <br />
 
-# 🎃 Hacktoberfest 🎃
+
 
 [All the details you need](https://github.com/OperationCode/START_HERE/blob/master/README.md#-hacktoberfest-) before participating with us during Hacktoberfest.
```
aaron-suarez, line 11: We can remove this line as well

aaron-suarez:
> I think since members participated instead of deleting we should run a series of short articles highlighting those who is did (why, how they got involved, etc) and linking to their GH and LI accounts.

@C-Lion if you'd like to do this, feel free to open up a new PR. But I think the banner can be removed from the README because the event is indeed over.

PR #551, opened 2022-06-21
Adds flask admin as a feature. Bumps flask admin to 1.6.0

oafernandes:
-Adds flask admin feature
-Bumps flask admin to 1.6.0 and all dependencies
-Issue with password hashing when trying to add a new admin from admin view, password is not hashed and stored as plaintext.


.env
```diff
@@ -7,3 +7,7 @@ FLASK_DEBUG=1
 ALGOLIA_APP_ID=search_id
 ALGOLIA_API_KEY=search_key
 INDEX_NAME=resources_api
+SECRET_KEY=sammy
+SECURITY_PASSWORD_SALT=saltedpop
+ADMIN_EMAIL=test@me.com
+ADMIN_PASSWORD=1234
```
aaron-junot, line 13: Can we please add a newline to the end of this file?

Dockerfile
```diff
@@ -19,6 +19,8 @@ RUN apt-get update \
     && pip install poetry \
     && poetry config virtualenvs.create false
 
+RUN poetry lock
```
aaron-junot, line 22: Why is this necessary?

configs.py
```diff
@@ -37,6 +37,13 @@ def get_sys_exec_root_or_drive():
 if not all([algolia_app_id, algolia_api_key]):
     print("Application requires 'ALGOLIA_APP_ID' and 'ALGOLIA_API_KEY' for search")
 
+secret_key = os.environ.get('SECRET_KEY', None)
+security_password_hash = 'pbkdf2_sha512'
```
aaron-junot, line 41: Can we make this an environment variable?

configs.py
```diff
@@ -37,6 +37,13 @@ def get_sys_exec_root_or_drive():
 if not all([algolia_app_id, algolia_api_key]):
     print("Application requires 'ALGOLIA_APP_ID' and 'ALGOLIA_API_KEY' for search")
 
+secret_key = os.environ.get('SECRET_KEY', None)
```
aaron-junot, line 40:
```suggestion
secret_key = os.environ['SECRET_KEY']
```

configs.py
```diff
@@ -37,6 +37,13 @@ def get_sys_exec_root_or_drive():
...
+security_password_hash = 'pbkdf2_sha512'
+security_password_salt = os.environ.get('SECURITY_PASSWORD_SALT', None)
+
+if not all([secret_key, security_password_salt]):
+    print('Application requires "SECRET_KEY" and "SECURITY_HASH"')
```
aaron-junot, line 45: We can remove this if we use the bracket syntax instead of `.get()`

run.py
```diff
@@ -15,6 +25,49 @@
 })
 
 
+# @event.listens_for(User.password, 'set', retval=True)
+# def hash_user_password(target, value, oldvalue, initiator):
+#     """Encrypts password when new admin created in User View"""
+#     if value != oldvalue:
+#         return utils.encrypt_password(value)
+#     return value
```
aaron-junot, line 33: What is this?

run.py
```diff
@@ -15,6 +25,49 @@
...
+
 @app.shell_context_processor
 def make_shell_context():
-    return {'db': db, 'Resource': Resource, 'Category': Category, 'Language': Language}
+    return {'db': db, 'Resource': Resource, 'Category': Category, 'Language': Language,
+            'User': User, 'Role': Role}
+
+
+@app.before_first_request
+def before_first_request():
```
aaron-junot, line 53: Please rename this so we don't accidentally shadow or get confused. Choose a name that is descriptive of what the function is doing

run.py
```diff
@@ -15,6 +25,49 @@
...
+    """ Adds admin/user roles and default admin account and password if none exists"""
+    db.create_all()
```
aaron-junot, line 55: Is this idempotent?

run.py
```diff
@@ -15,6 +25,49 @@
...
+    user_datastore.find_or_create_role(name='admin', description='Administrator')
+    user_datastore.find_or_create_role(name='user', description='End User')
+
+    admin_email = os.environ.get('ADMIN_EMAIL', "admin@example.com")
+    admin_password = os.environ.get('ADMIN_PASSWORD', 'password')
```
aaron-junot, line 60: We want this to blow up. Please use bracket syntax here

run.py
aaron-junot, line 53: If the admin user already exists when this function runs, what happens? Are there side effects?

aaron-junot (changes requested):

Dockerfile
oafernandes: I manually added the dependency versions to the pyproject.toml file. Below link to solution for resolving the "....which doesn't match any versions, version solving failed." error

[Known issue]https://github.com/python-poetry/poetry/issues/1281#issuecomment-534583940

.env
oafernandes: Added

configs.py
oafernandes: Uses bracket notation instead of .get()

configs.py
oafernandes: SECURITY_PASSWORD_HASH added to .env

configs.py
oafernandes: Uses bracket notation instead of .get()

run.py
oafernandes: This listens for when a new admin password is added in the admin view and encrypts with same method as before_first_request(). Removed.

run.py
oafernandes: changed to add_admin_role

run.py
oafernandes: Changed to bracket from `.get()`

run.py
oafernandes: The only way to check would be to delete the table from the database I think? May need help with this.

PR #586, opened 2023-12-10
chore(#82): Consistent Error Presentation

danielburn5:
# Intro
👋 This is my first PR to this project! It's highly likely I have misunderstood something along the way. Happy to revisit and rework any of this!

# What is this?

I started looking at some of the complexity issues raised by Code Climate for #82, specifically that of `app.api.validations.requires_body`.

I realised that the cognitive complexity can be reduced by removing the explicit error handling in this function and letting flask raise the appropriate error (see [Request.get_json](https://tedboy.github.io/flask/generated/generated/flask.Request.get_json.html)) if the JSON is malformed (and there is already an error handler for this). I think this code only needs to handle the error case which is specific to the app (ie. we must have content in the payload).

However, this uncovered inconsistencies in the formatting of API responses when it comes to errors. It caused a lot of test failures and a lot of head scratching on my part.

It looks as though the codebase already attempts to standardise responses, but as the errors payload is sometimes passed in, it has the potential to be inconsistent. Sometimes it is an array of dicts, sometimes just a single dict.

Long term, I think the creation of error responses could be standardised further but for now I updated each of the manual entries so that errors are always presented as arrays of dicts, like this:

```
{
    "errors": [
        {
            "message": ""
        }
    ]
```

# Testing
Updated existing tests in line with changes.

# Question
Might these changes to error responses have any effect on the front end?

### DIFF
--- a/app/api/validations.py
+++ b/app/api/validations.py
@@ -11,15 +11,11 @@ INVALID_TYPE = "invalid-type"
 
 def requires_body(func):
     def wrapper(*args, **kwargs):
-        try:
-            # JSON body is {} or []
-            if not request.get_json():
-                return missing_json_error()
-        except Exception as e:
-            # JSON body is completely missing
-            if "Expecting value: line 1 column 1" in str(e):
-                return missing_json_error()
-
+        # JSON body is {} or []
+        # This will throw a 400 if the JSON is malformed
+        # and drop into handlers.bad_request()
+        if not request.get_json():
+            return missing_json_error()
         return func(*args, **kwargs)
     return wrapper
 
--- a/app/api/validations.py
+++ b/app/api/validations.py
@@ -1,4 +1,5 @@
 from flask import request
+from werkzeug.exceptions import BadRequest
 from app.models import Resource
 from app.utils import standardize_response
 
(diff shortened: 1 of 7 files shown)


tests/unit/test_routes/test_invalid_put_post.py
```diff
@@ -97,15 +97,15 @@ def test_validate_resource(module_client, module_db, fake_auth_from_oc):
                           )
     assert_missing_body(response)
 
-    # Data cannot be empty
+    # Data must be valid json
```
danielburn5, line 100: This test was not quite working as intended previously. Passing in `data=''` is not the same as not passing in `data` at all. This ultimately gets treated as a 400 bad request as there is a payload, but it's not valid json.

app/api/validations.py
```diff
@@ -10,23 +10,19 @@
...
-        except Exception as e:
-            # JSON body is completely missing
-            if "Expecting value: line 1 column 1" in str(e):
-                return missing_json_error()
-
+        # JSON body is {} or []
+        # This will throw a 400 if the JSON is malformed
+        # and drop into handlers.bad_request()
+        if not request.get_json():
+            return missing_json_error()
```
danielburn5, line 17: @aaron-junot Is there a way to check if this has had the desired effect on the cognitive complexity in Code Climate whilst I am on a fork?

danielburn5:
Closing. Didn't realise this repo was decomissioned.

